PR #16, opened 2023-02-07
Inclusion of token authenticity

gklps:


core/token_auth.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"io/ioutil"
+	"strconv"
+	"strings"
+)
+
+func (c *Core) getWholeTokenValue(wholeTokenName string) (int, string) {
```
muraliens, line 11: Add this as separate package outside core, like RAC create separate package token and add this files,

core/token_auth.go
```diff
@@ -0,0 +1,93 @@
...
+	response.Close()
+
+	tokenLevel := string(trimmedResult[:len(trimmedResult)-64])
+	tokenCountHash := string(trimmedResult[len(trimmedResult)-64:])
+
+	fmt.Println(len(tokenCountHash), "token len")
+
+	tokenLevelInt, err := strconv.Atoi(tokenLevel)
+	if err != nil {
+		fmt.Println("Error occured", "error", err)
```
muraliens, line 36: handle the errors properly

gklps:
@muraliens Made the changes as per your comments

core/token_auth.go
gklps: moved to a separate package called token

core/token_auth.go
gklps: updated

muraliens (approved):

Issue #48, opened 2023-05-11
Receiver offline ; quorum pin tokenstate


harirubix:
Tokenstate is essentially { tokenhash.tokenchainhash } at time when the current sender received a token. Now, when the sender initiates a transaction, the quorum for this transaction pins tokenstate to mark that the ‘tokenstate’ is ‘exhausted’. If a double spend attempt is initiated for the same token , the new quorum will check for token state in ipfs and therefore block double spend attempt

![Receiver_offline_quorum_pin_tokenstate](https://github.com/rubixchain/rubixgoplatform/assets/130964671/06745d26-7cf9-441b-8f06-fcf87b24b65e)

KiranHRubix:
@harirubix 

In this logic we also have to consider the fact that the latest blocks will be self signed blocks,

Like Pledged BLock, Un pledged Block,

KiranHRubix:
I will be handling the scenarios in code, But just a heads up

KiranHRubix:
This issue was handled in PR #63

PR #52, opened 2023-05-24
Feature security tokenstatepincheck

KiranHRubix:
This PR is related to Issue #48


core/quorum_recv.go
```diff
@@ -142,6 +142,42 @@ func (c *Core) quorumRBTConsensus(req *ensweb.Request, did string, qdc didcrypto
...
+	}
+	wg.Wait()
+
+	for i := range tokenStateCheckResult {
+		if tokenStateCheckResult[i].Error != nil {
+			c.log.Error("Error occured", "error", err)
+			crep.Message = "Error while cheking Token State Message : " + tokenStateCheckResult[i].Message
+			return c.l.RenderJSON(req, &crep, http.StatusOK)
+		}
+		if tokenStateCheckResult[i].Status {
```
vyshnavsdeepak, line 170: Does status true mean state exhausted?

core/quorum_validation.go
```diff
@@ -2,15 +2,26 @@ package core
...
 	"github.com/rubixchain/rubixgoplatform/contract"
+	"github.com/rubixchain/rubixgoplatform/core/wallet"
 	"github.com/rubixchain/rubixgoplatform/did"
 	"github.com/rubixchain/rubixgoplatform/token"
 	"github.com/rubixchain/rubixgoplatform/util"
 )
 
+type TokenStateCheckResult struct {
+	Token                 string
+	Status                bool
```
vyshnavsdeepak, line 19: shall we change this bool to enum/const of known statuses?

core/quorum_validation.go
```diff
@@ -2,15 +2,33 @@ package core
...
+	Error                 error
+	Message               string
+	tokenIDTokenStateData string
+}
+
+const (
+	exhausted = true
```
vyshnavsdeepak, line 26: it doesn't make sense to have enum if the only two values are true and false.

vyshnavsdeepak: I suggested enum to have more than 2 statuses

vyshnavsdeepak: Exhausted sounds negative, and true being its value is confusing.

core/quorum_validation.go
vyshnavsdeepak: Read 1 & 2,
3. If the only goal ever with TokenStateCheckResult is whether it is exhausted or not,
It will be more clear if we just have a key called Exhausted(bool) instead of a Status (bool)

core/quorum_validation.go
vyshnavsdeepak: If we have it as 
const (
    Exhausted Status = iota
    Free
)

(here Status becomes a type)
It would be easier to add more Statuses later

Refer - https://www.sohamkamani.com/golang/enums/#enum-types

core/quorum_validation.go
vyshnavsdeepak: I would still prefer to go with enum for future development compatibility.

KiranHRubix (commented):
Changes have been added based on the feedback

core/quorum_validation.go
```diff
@@ -2,15 +2,28 @@ package core
...
 
+type TokenStateCheckResult struct {
+	Token                 string
+	Exhausted             bool
+	Error                 error
+	Message               string
+	tokenIDTokenStateData string
+}
+
+const ()
```
vyshnavsdeepak, line 25: here

KiranHRubix (commented):
removed unused const

core/quorum_validation.go
```diff
@@ -167,3 +178,117 @@ func (c *Core) getUnpledgeId(wt string) string {
...
+			result.Message = "Error fetching block Id"
+			resultArray[index] = result
+		}
+		//concat tokenId and BlockID
+		tokenIDTokenStateData := tokenId + blockId
+		tokenIDTokenStateBuffer := bytes.NewBuffer([]byte(tokenIDTokenStateData))
+
+		//add to ipfs get only the hash of the token+tokenstate
+		tokenIDTokenStateHash, err := c.ipfs.Add(tokenIDTokenStateBuffer, ipfsnode.Pin(false), ipfsnode.OnlyHash(true))
+		if err != nil {
```
vyshnavsdeepak, line 211: We can simplify the logic we use empty returns at each error after updating resultArray

KiranHRubix: REsolved

KiranHRubix (commented):
chnages accomadated from feedback

core/quorum_validation.go
```diff
@@ -167,3 +178,98 @@ func (c *Core) getUnpledgeId(wt string) string {
...
+func (c *Core) pinTokenState(tokenStateCheckResult []TokenStateCheckResult, did string) {
+	for i := range tokenStateCheckResult {
+		tokenIDTokenStateBuffer := bytes.NewBuffer([]byte(tokenStateCheckResult[i].tokenIDTokenStateData))
+		id, err := c.w.Add(tokenIDTokenStateBuffer, did, wallet.QuorumRole)
+		if err != nil {
+			c.log.Error("Error triggered while adding token state", err)
+		}
+		_, err1 := c.w.Pin(id, wallet.QuorumRole, did)
+		if err1 != nil {
+			c.log.Error("Error triggered while pinning token state", err)
```
vyshnavsdeepak, line 271: need return at the end of these error checks as well, right?

KiranHRubix (commented):
resolved

vyshnavsdeepak (approved):

KiranHRubix (commented):
resolved

core/quorum_validation.go
```diff
@@ -167,3 +178,110 @@ func (c *Core) getUnpledgeId(wt string) string {
 	}
 	return b.GetUnpledgeId()
 }
+
+/*
+ * Function to check whether the TokenState is pinned or not
+ * Input tokenId, index, resultArray, waitgroup,quorumList
+ */
+func (c *Core) checkTokenState(tokenId, did string, index int, resultArray []TokenStateCheckResult, wg *sync.WaitGroup, quorumList []string) {
```
harirubix, line 186: Would it be a good option to enable this check at receiver side also? It's more a precautionary action by receiver to make sure the DHTprovs nodes are only quorum for its own transaction. In worst case situations if ever 2 consensus go through at same time , it can be detected at receiver level.

harirubix (approved):

KiranHRubix (commented):
changes reviewed

harirubix:
@muraliens please merge this PR

KiranHRubix:
@harirubix , I am rebasing the code to the curretn develop branch so that conflicts while merging will not be there, I will update by end of day , post which can be merged

KiranHRubix:
@harirubix, The refactor is complete, 

But in this We still have not come up with a logic to release the tokenstate pin ,

Issue #47, opened 2023-05-10
Txn Fail due to receiver offline - Bug 


KiranHRubix:
- During transfer from node 1 to node 2 after the consensus finishes successfully, the token details are shared to Receiver to verify
- During this time if at any case the receiver side fails due to IPFS port killed issue or the executable shuts down due to panic , the transfer fails
- But since consensus passed the token chain has been updated for the new block
- Trying to transfer the tokens again, it will trigger latest block cannot be synced error and hence the token ownership check will not pass

neerajchowdary889:
Hello sir, What is understood is that node 1’s transaction completed its consensus and details were shared to verify. At this time, there was a transfer failure at receiver end and the transaction failed. But when node 1 tried to initiate the transaction again, the receiver node was not in sync with the updated token chain, so an error is raised.
Is it correct what I understood?

KiranHRubix:
yes

neerajchowdary889:
Can we use state sync or checkpoint sync to get the node synchronized with the latest token chain? Not only in this case, but we can also resync other nodes that are out of sync, right? I’m not that sure, but these are my thoughts about it.

KiranHRubix:
Fix for the issue has been implemented in following way
- If the said issue arises after consensus a status of Txn Finality Pending will be added in the a DB table of pending txn to the associated txn id with the related details.
- Furthermore the token chain would be synced from the quorum to the sender.
- Once the receiver comes back online after resolving the issue, a Txn finality command for the specific txn ID would be triggered where the rest of the Validation on receiver happens and the txn fin laity is achieved.

KiranHRubix:
The Issue has been resolved , By adding the Tokenstate information to the quorum and then allowing the quorum to sync only after the receiver get the Tokens . Thus always checking if receiver node is online for the transaction

PR #55, opened 2023-06-02
Bug Fix - Txn Finality if Receiver Offline

KiranHRubix:
This PR is related to #47 

List of changes/additions

struct  update
check block value is nil
update based on txn_id
 txn status update after consensus
txnstatus updation logic modification
added loggers
errors handled
sync/pin token & chain info - receiver  finality
removed dead code
modified add and update txn status method
uopdated tokenchain sync on receiver
added debug logs
minor update
fixed sender side finality updation

### DIFF
--- a/client/txn_finality.go
+++ b/client/txn_finality.go
@@ -1,0 +1,26 @@
+package client
+
+import (
+	"github.com/rubixchain/rubixgoplatform/core/model"
+	"github.com/rubixchain/rubixgoplatform/server"
+)
+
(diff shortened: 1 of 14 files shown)


KiranHRubix:
The PR is related to issue where if in case after a txn isn intitated and receivergoes foffline after consensus passes , txn fails. The Logic/Code solution added saves such txn in a queue in table and once the Receiver is back online, initiated for finality of such failed txn

KiranHRubix:
PR is outdated closing this and opening new PR

PR #58, opened 2023-06-08
Feature/type1 quorum selection

### LINKED ISSUE #42: Quorum selection
Quorum selection can be two ways

1. Network selection of quorum ( open network , previously type 1) : Based on transaction id.
2. Pick your own quorum ( closed network , previously type 2)

ashi31:
closes #42 
This PR is to select the type 1 quorums from DIDPeerTable where quorums are selected based on the last character of transaction ID should match with last character of Quorum DID.


core/quorum_initiator.go
```diff
@@ -440,9 +449,15 @@ func (c *Core) finishConsensus(id string, qt int, p *ipfsport.Peer, status bool,
 	}
 }
 
-func (c *Core) connectQuorum(cr *ConensusRequest, addr string, qt int) {
+func (c *Core) connectQuorum(cr *ConensusRequest, addr string, qt int, pledgetokens float64) {
 	c.startConsensus(cr.ReqID, qt)
-	p, err := c.getPeer(addr)
+	var p *ipfsport.Peer
+	var err error
+	if cr.Type == 1 {
```
harirubix, line 456: Seems you are doing the same operation in both if and else blocks.

core/quorum_initiator.go
```diff
@@ -440,9 +449,15 @@ func (c *Core) finishConsensus(id string, qt int, p *ipfsport.Peer, status bool,
 	}
 }
 
-func (c *Core) connectQuorum(cr *ConensusRequest, addr string, qt int) {
+func (c *Core) connectQuorum(cr *ConensusRequest, addr string, qt int, pledgetokens float64) {
```
harirubix, line 452: `pledgetokens` is not used anywhere . Also for later , Please use camelCase while naming variables

core/quorum.go
```diff
@@ -47,10 +50,35 @@ func NewQuorumManager(s storage.Storage, log logger.Logger) (*QuorumManager, err
...
+			qm.log.Error("Quorums not present")
+			return nil
+		}
+		if len(quorumList) < 5 {
+			qm.log.Error("Not enough quorums present")
+			return nil
+		}
+		var quorumAddrList []string
+		fmt.Println(quorumAddrList)
+		for _, q := range quorumList {
```
harirubix, line 77: If the `quorumAddrList` returned is greater than 7 , Wouldn't the sender connect to all available quorum instead of upper limit 7? Either the append operation can be stopped once `quorumAddrList` length reaches 7 or handled at `initiateConsensus()` by connecting to quorum and checking for balance.

thewebchap (changes requested):

KiranHRubix:
@ashi31 Kindly go through the comments and resolve the conflicts so that it wont be a blocker for code merging

core/quorum_initiator.go
ashi31: Removed the if else blocks

core/quorum_initiator.go
ashi31: Removed the variable pledgetokens

core/quorum.go
ashi31: Yes, I agree. The sender will try to connect to all the available quorums.
If we stop the append operation once the length is reached, we might have quorums that are offline. In this case, we'll have to get the list again.
After we have the list of quorums, we are trying to connect to the quorums and pledge tokens. But yes, a check for balance would be important. In initiateConsensus(), we can put a check probably where we are trying to connect with quorums, I would appreciate if we can discuss this once before making any changes. 
Another thing related to quorum count, from what I understand, there are no checks for type2 quorums as well, do we need to check that as well?

harirubix: Is this resolved ? @ashi31  . Right now the sender will try to connect to all nodes satisfying the condition. 
Ideally You could do a generic ping check to make sure nodes are alive( This has to be done for both type 1 and type2) 
or as temporary fix only pick 7.

thewebchap (changes requested):

core/quorum.go
ashi31: Hi, The temporary fix to select 7 quorums is added. And the datatoken file is reverted back to original.

thewebchap (approved):

core/wallet/ballance.go
```diff
@@ -0,0 +1,22 @@
+package wallet
+
+import "fmt"
+
+//TODO:Change function for part tokens
+
+func (w *Wallet) GetBalance(did string, peerID string) float64 {
```
vaishnav-ch, line 7: I think this function is not using anywhere.

vaishnav-ch (approved):

PR #60, opened 2023-07-06
Periodic Pledging

aparnaullas-rubix:
#60 
### A Draft PR that involves the working logic for Periodic Pledging (Un-pledging part)

Following are the major sub-points to keep a track of:

1. Added epoch time to token chains and transaction history
2. Added check for epoch time 24h elapsed (replacement check for PoW)
3. Unpledging based on periodic pledging 
4. 8 hourly background check for unpledging after 24h

_Pending logic to be implemented: Re-pledhing (Self-Transfer) for the tokens in the unsafe state_

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -51,6 +51,7 @@ const (
 	SetupQuorumCmd        string = "setupquorum"
 	GenerateTestRBTCmd    string = "generatetestrbt"
 	TransferRBTCmd        string = "transferrbt"
+	SelfTransferRBTCmd    string = "selftransferrbt"
 	GetAccountInfoCmd     string = "getaccountinfo"
 	SetupServiceCmd       string = "setupservice"
 	DumpTokenChainCmd     string = "dumptokenchain"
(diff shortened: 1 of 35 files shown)


command/command.go
gklps, line 1: This was a debug statement right? Lets remove this?

aparnaullas-rubix:
#60  
### This pull request involves the working logic for Self-Transfer

Following are the major sub-points to keep a track of:

1. Added the logic for self-transfer to get new nodes/tokens to pledge for the unsafe tokens (re-pledge logic)
2. Added the API call "selftransferrbt"
3. Changed the time check to 16 days for unpleding and self transfer (can be changed when the time frame is confirmed)
4. All checks are down 8 hourly (can be changed when the time frame is confirmed)

_Note: There was check in the quorum_receive function (to which I made no changes to) but all of a sudden it started throwing errors. So For this PR, I have commented that section which can be discussed and sorted_

PR #73, opened 2023-10-25
adding and removing explorer URLs to config file

nidhinmahesh:
Introduced two new APIs to add and remove explorer URLs dynamically. These changes were essential since the `api_config.json` file is encrypted, making external modifications impossible. Also updated the `Core` struct to handle an array of explorer URLs.

### Changes
- [x] Added API to add explorer URL.
- [x] Added API to remove explorer URL.
- [ ] Updated `Core` struct to accept an array of explorer URLs .
- [ ] Run batch tests to validate all explorer URLs.

### Motivation
1. Encrypted `api_config.json` prevents any external modifications for security reasons.
2. The need for dynamically updating explorer URLs without compromising security.
3. Keeping changes centralized and secure through APIs.

### Testing
Batch tests will be run to ensure all explorer URLs are functional.

### DIFF
--- a/client/explorer.go
+++ b/client/explorer.go
@@ -1,0 +1,39 @@
+package client
+
+import (
+	"github.com/rubixchain/rubixgoplatform/core/model"
+	"github.com/rubixchain/rubixgoplatform/setup"
+)
+
+func (c *Client) AddExplorer(links []string) (string, bool) {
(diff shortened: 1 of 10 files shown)


harirubix:
duplicate to #92

PR #93, opened 2024-02-07
Fix/Parts token decimal point bug

vaishnav-ch:
1. `MaxDecimalPoint` constant added with a value of `3` and used for determining minimum transaction amount and pledge amount.
2. Minimum transaction amount and pledge amount is 0.001. So, If the transaction amount is less than 0.005 it is changed to 0.005 as total pledge amount for meeting the requirement of minimum pledge amount per quorum (0.001).
3. Length of the smart contract hash is verified before subscribing contract to avoid invalid subscriptions.


harirubix (approved):

vaishnav-ch:
Code changes in this PR is added in the PR `https://github.com/rubixchain/rubixgoplatform/pull/173`. So closing this PR without merging.

PR #96, opened 2024-02-09
deterministic ecdsa using go 1.19 legacy

### LINKED ISSUE #94: Use deterministic ECDSA key
In Go 1.20 the crypto/ecdsa.GenerateKey call uses MaybeReadByte, so now all GenerateKey implementation are non-deterministic. This hinders our option for key recovery. Although moving to BIP39 will solve this, an interim solution is to use Go 1.19 legacy package that creates a deterministic key with a user provided byte[] used as seed.

Refer [discussion](https://github.com/golang/go/issues/58637)

harirubix:
Closes #94 
Temporary feature
Uses `-privPWD` as seed to generate privatekey. This can be a good addition for 2 scenarios while testing simulations where we can create same set of pki pairs.  This is not to be used in production since the seed is essentially a string and nodes that use same seed will end up having same private keys. The production level solution to this is using BIP39 that has a higher entropy using mnemonics.

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -6,9 +6,13 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/sha512"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+
+	"filippo.io/keygen"
+	"golang.org/x/crypto/hkdf"
 )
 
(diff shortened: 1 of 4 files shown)


KiranHRubix (approved):
code changes looks good and informative. lets move to testing and then merge.

Maneesha-rubix (approved):

PR #101, opened 2024-03-04
Changed default DID mode

Maneesha-rubix:
Setting Basic mode as the default mode and Light mode as fourth mode while creating DID.

### DIFF
--- a/did/model.go
+++ b/did/model.go
@@ -4,8 +4,8 @@ const (
 	BasicDIDMode int = iota
 	StandardDIDMode
 	WalletDIDMode
-	LightDIDMode
 	ChildDIDMode
+	LightDIDMode
 )
 
 const (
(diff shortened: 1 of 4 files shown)


did/model.go
```diff
@@ -1,10 +1,10 @@
 package did
 
 const (
-	LightDIDMode int = iota
-	BasicDIDMode
+	BasicDIDMode int = iota
 	StandardDIDMode
 	WalletDIDMode
+	LightDIDMode
```
harirubix, line 7: Could have added as option 4 rather than 3 . It works fine now since we dont usually make wallets with child. But this might open some corner case issues during switch or if/else statements.

harirubix (approved):
Code looks fine.  Please test different scenarios before merging

did/model.go
Maneesha-rubix: In that case I will change the sequence again.

vaishnav-ch (approved):

harirubix (approved):

PR #136, opened 2024-04-18
Fixed error occurring while trying to transfer RBT with fractional value component, but sender only possessing sufficient whole RBT tokens

arnabghose997:
## Issue

Suppose the sender has 2 whole RBT tokens. The error occurred when they attempted to send 1.5 RBT.

Error Log:

```
2024-04-18T17:34:43.917+0530 [ERROR] Main.Core.wallet: Failed to get tokens: err="no records found"
2024-04-18T17:34:43.917+0530 [ERROR] Main.Core: No part Tokens found , This wallet is empty: err="no records found"
2024-04-18T17:34:43.918+0530 [ERROR] Main.Core: Failed to get tokens: err="no records found"
```

Upon analysis, it was observed that the function `GetAllPartTokens` was throwing back an error if there was no part tokens found in `TokensTable`. However, in case of transferring 1.5 RBT, the sender already has ample balance for a successful transfer and hence it should not throw an error

## Fix

While handling error from `GetAllPartTokens`, we check if the error string contains the value `no records found`. If yes, then the `requiredToken` and `reqAmt` are returned without any error. If no, then an error is thrown. In this way, we deal with `reqAmt` in later stages where the creation of Part tokens occurs.

### DIFF
--- a/core/token.go
+++ b/core/token.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/rubixchain/rubixgoplatform/block"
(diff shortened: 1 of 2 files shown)


KiranHRubix (approved):
Logic looks good , Procceed to test and if testing done lets go to merge

gklps (approved):

PR #140, opened 2024-04-23
Feat/allen/pincheck

Allen-Cherian:
Changes Made:

Struct Update: Added new fields (TransactionId, sender, receiver, and token value) to the existing TokenProviderMap struct. These additions facilitate the population of the TokenProviderTable. The updates to TokenProviderMap enhance the information stored in TokenProviderTable, allowing for better tracking and management of transactions.

Quorum-based Pinning: Implemented pinning of the parent token by the quorum, while transferring the part token.

### DIFF
--- a/core/quorum_validation.go
+++ b/core/quorum_validation.go
@@ -393,7 +393,7 @@ func (c *Core) checkTokenState(tokenId, did string, index int, resultArray []Tok
 	resultArray[index] = result
 }
 
-func (c *Core) pinTokenState(tokenStateCheckResult []TokenStateCheckResult, did string) error {
+func (c *Core) pinTokenState(tokenStateCheckResult []TokenStateCheckResult, did string, transactionId string, sender string, receiver string) error {
 	var ids []string
 	for i := range tokenStateCheckResult {
 		tokenIDTokenStateBuffer := bytes.NewBuffer([]byte(tokenStateCheckResult[i].tokenIDTokenStateData))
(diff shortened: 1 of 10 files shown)


core/quorum_validation.go
```diff
@@ -175,6 +177,11 @@ func (c *Core) validateTokenOwnership(cr *ConensusRequest, sc *contract.Contract
 				c.log.Error("failed to sync parent token chain", "token", pt)
 				return false, err
 			}
+			_, err = c.w.Pin(pt, wallet.WholeTokenSplitLockRole, quorumDID, cr.TransactionID, address, receiverAddress, ti[i].TokenValue)
```
KiranHRubix, line 180: how are we identifying if the parent token is a whole token or part token , when giving the role WholeTokenSplitLockRole

KiranHRubix (commented):
Code chan ges looked good , Please Document the testing and the results you are ssing after testing

KiranHRubix (approved):

core/quorum_validation.go
Allen-Cherian: So in this case we are not identifying whether the parent token is a whole token or part token. We are just pinning the parent token from the quorum. So may be I can change the role to quorum role instead of wholeTokenSplitLockRole ?

KiranHRubix: Maybe create a Role like "ParentTokenPinning" this would be a bit suitable for the operation we are doing

KiranHRubix (approved):
chnages looks good , if testing done we can go to merge

PR #141, opened 2024-04-23
7 quorums fix

ashi31:
Selection of 5 quorums out of 7 quorums. The first 5 quorums including the pledging quorum will be in the signers list.

The below cases were tested : 

All 7 quorums set up 
Out of 7 quorums, only 5 were set up

In above cases, the pledging quorum completed consensus in the first 4 runs and signed after the first 4th run as well.


core/quorum_initiator.go
```diff
@@ -656,29 +658,53 @@ func (c *Core) finishConsensus(id string, qt int, p *ipfsport.Peer, status bool,
...
+		if p != nil {
+			p.Close()
+		}
+		return
+	}
+	keys := make([]string, 0, len(pd.PledgedTokens))
+	for k := range pd.PledgedTokens {
+		keys = append(keys, k)
+	}
+	key := keys[0]
```
harirubix, line 679: Can you rename variable `key` into a more clearer name ? something like PledgedTokenDID for eg.  Also is the for loop required ?

thewebchap (changes requested):

core/quorum_initiator.go
```diff
@@ -656,29 +658,53 @@ func (c *Core) finishConsensus(id string, qt int, p *ipfsport.Peer, status bool,
...
 		if status {
 			did := p.GetPeerDID()
-			if cs.Result.SuccessCount < MinConsensusRequired {
-				csig := CreditSignature{
-					Signature:     util.HexToStr(ss),
-					PrivSignature: util.HexToStr(ps),
-					DID:           did,
-					Hash:          hash,
-				}
+			csig := CreditSignature{
```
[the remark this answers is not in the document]
ashi31: Right, passing the length of quorum list, might work here. So it will be like else if cs.Result.RunsTotal == 6 && cs.Result.SuccessCount == MinConsensusRequired-1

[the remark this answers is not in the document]
harirubix: I am not sure. Please check.

This is the scenario I am talking about. 

QuorumList : Q1 , Q2 , Q3 , Q4 , Q5, Q6 , Q7

Pledged Quorum : Q1

Initaiate consensus : Q1 , Q2 , Q3 , Q4 respond .   Now say Q5 and Q6 systems are not responding , when you get respond from Q7 , you won't accept it with check `else if cs.Result.RunsTotal == len(quorumList)-1` since RunsTotal is 4 now.

core/quorum_initiator.go
```diff
@@ -656,29 +666,55 @@ func (c *Core) finishConsensus(id string, qt int, p *ipfsport.Peer, status bool,
...
+			}
+			if cs.Result.SuccessCount < MinConsensusRequired-1 {
+				cs.P[did] = p
+				cs.Credit.Credit = append(cs.Credit.Credit, csig)
+				cs.Result.SuccessCount++
+			} else if did == pledgingDID && cs.Result.SuccessCount == MinConsensusRequired-1 {
+				cs.P[did] = p
+				cs.Credit.Credit = append(cs.Credit.Credit, csig)
+				cs.Result.SuccessCount++
+			} else if cs.Result.RunningCount == 0 && cs.Result.SuccessCount == MinConsensusRequired-1 {
```
harirubix, line 711: In this case , with 7 nodes setup and running , aren't you counting vote of last (7th) node to respond instead of the 5th node?  5th and 6th nodes response will be discarded since RunningCount will not be 0.  Hence waiting time is more and early responders don't get credit.

thewebchap (changes requested):

core/quorum_initiator.go
ashi31: This can be resolved. For the above scenario, we would need to perform a few more checks and probably need one more variable. So, that is a tradeoff here. Also, since the process would get more complicated when equal pledging is implemented with type1 quorums and the above algorithm will change for type1. And in type2, we are using our own quorums and order of their signature will not be fixed, so I was thinking if we can bear this tradeoff. Please let me know, if the changes need to be made, I will do it right away.

core/quorum_initiator.go
```diff
@@ -150,6 +155,7 @@ func (c *Core) QuroumSetup() {
 	c.l.AddRoute(APIUpdatePledgeToken, "POST", c.updatePledgeToken)
 	c.l.AddRoute(APISignatureRequest, "POST", c.signatureRequest)
 	c.l.AddRoute(APISendReceiverToken, "POST", c.updateReceiverToken)
+	c.l.AddRoute(APIUnlockTokens, "POST", c.unlockTokens)
```
gklps, line 158: Will this unlock all the locked token or just the token locked in this specific pledge request?

ashi31: It will unlock the tokens for the specific pledge request.

core/quorum_initiator.go
ashi31: We can maybe create a list of dids that pledged and check if the signing did is on that list or not. Or maybe a map that will hold the pledging quorum did, with did as the key and value will be a boolean value. Whenever a pledging did signs, we can update the boolean value to True. Using a list would increase the number of iterations. I think a map would be better in this case.

thewebchap (approved):

PR #142, opened 2024-04-24
type1 quorum transactionID updated

ashi31:
The last characters we get in DID are : a, e, i, m, q, u, y, 4. 
Transaction IDs for Type 1 transaction are added into IPFS, to get the relative IPFS ID so that the last character matches with Quorum DIDs.

### DIFF
--- a/core/quorum_initiator.go
+++ b/core/quorum_initiator.go
@@ -1,14 +1,19 @@
 package core
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	files "github.com/ipfs/go-ipfs-files"
 	"github.com/rubixchain/rubixgoplatform/block"
 	"github.com/rubixchain/rubixgoplatform/contract"
 	"github.com/rubixchain/rubixgoplatform/core/ipfsport"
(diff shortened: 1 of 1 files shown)


ashi31:
The Old PR is removed.

harirubix:
This works ! But I am sure we can come with a better solution.  Could you please check the multihash structure for IPFS ? That will bring an alt options.

ashi31:
> This works ! But I am sure we can come with a better solution. Could you please check the multihash structure for IPFS ? That will bring an alt options.

Yes sure Hari, I'll check that

core/quorum_initiator.go
```diff
@@ -1245,3 +1257,63 @@ func (c *Core) createCommitedTokensBlock(newBlock *block.Block, smartContractTok
...
 	return nil
 }
+
+func (c *Core) generateTransactionID(transType int, sc *contract.Contract) (string, string, error) {
+	hash := util.CalculateHash(sc.GetBlock(), "SHA3-256")
+	tid := util.HexToStr(hash)
+	if transType == QuorumTypeTwo {
+		return tid, "", nil
+	}
+	os.Mkdir(c.didDir+"/tempdir", 0777)
```
gklps, line 1267: What happens if the directory already exists?

harirubix:
> > This works ! But I am sure we can come with a better solution. Could you please check the multihash structure for IPFS ? That will bring an alt options.
> 
> Yes sure Hari, I'll check that

Any updates ?

ashi31:
> > > This works ! But I am sure we can come with a better solution. Could you please check the multihash structure for IPFS ? That will bring an alt options.
> > 
> > 
> > Yes sure Hari, I'll check that
> 
> Any updates ?

We can convert the Hex Transaction ID to Base32 temporarily, and use the last character of TransactionID to match with 2nd last character of quorum DID to pick quorums. That way we'll have more buckets to select quorum from.

harirubix:
Wontfix . Issue resolved by #177

PR #143, opened 2024-04-29
Fix: Smart contract rename

Allen-Cherian:
While working on smart contracts in WSL(Windows Subsystem for Linux), we came across an error which does not allow the renaming of a Temporary folder which is created while generating smart contract.
Fix :
The fix which is done here is that, we have written a function moveFile, which first tries to rename the folder, if an error pops up while renaming, the function tries to copy the file and generate the contract.

### DIFF
--- a/server/smart_contract.go
+++ b/server/smart_contract.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


gklps (approved):

gklps:
lg2me

PR #146, opened 2024-05-02
Add feature: Alternate quorum selection

Allen-Cherian:
This PR includes the feature for the addition of more than 1 set of quorums in a type 2 transaction. During the transaction the code will check whether the first set of quorums are available, if the first set of quorums are not available then the code will check for the next set of quorums and will resume the transaction.
Scenarios which have been tested :
1. First set of quorum not setup and only the second set setup : Transaction Successful with second set.
2. First set of quorum setup and the second one not setup : Transaction successful with first set.
3. Both the sets setup : Transaction successful with first set.

### DIFF
--- a/core/diagnostic.go
+++ b/core/diagnostic.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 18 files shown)


core/diagnostic.go
```diff
@@ -205,3 +207,59 @@ func (c *Core) ReleaseAllLockedTokens() model.BasicResponse {
 	response.Message = "All Locked Tokens Releases Successfully Or NO Locked Tokens to release"
 	return *response
 }
+
+func (c *Core) GetFinalQuorumList(ql []string) ([]string, error) {
+	// Initialize finalQl as an empty slice to store the groups that meet the condition
+	var finalQl []string
+	var opError error
+	// Loop through ql in groups of 5 items
+	for i := 0; i < len(ql); i += 5 {
```
harirubix, line 216: Shouldn't it be a list of 7?

harirubix (changes requested):

core/diagnostic.go
gklps: solved

gklps (approved):

gklps (approved):

harirubix (approved):

PR #173, opened 2024-06-05
Distributed pledging

vaishnav-ch:
- The transfer amount is equally divided with respect to the number of quorums participating in the transaction. 
Eg: If Transfer amount = 2.56789 then Pledge amount per quorum = 2.56789/5 = 0.51358 ( 0.513578 rounded )
- Minimum transaction amount and pledge amount is 0.00001. So, If the transaction amount is less than 0.00005 it is changed to 0.00005 as total pledge amount for meeting the requirement of minimum pledge amount per quorum (0.00001).


harirubix (approved):

harirubix (commented):
Not working in case of 7 quorums where one or two of 5 quorums doesnt have enough balance to pledge.

vaishnav-ch:
> Not working in case of 7 quorums where one or two of 5 quorums doesnt have enough balance to pledge.

This issue is resolved.

The reason behind this issue was that the function `c.GetFinalQuorumList()` was returning the quorum list with only 5 quorums even if the input quorum list have 7 quorums.

gklps:
LGTM. Pls attach testing results

PR #174, opened 2024-06-06
Quick fix : Mnemonic key error handling

harirubix:
### Small fix :  
Which we create type 4 DID , we check whether mnemonicFile is present (default -mnemonic.txt) unless specified using -mnemonicKeyFile flag.  We had a debug statement `Main.Core: failed to read mnemonic file: err="open mnemonic.txt: no such file or directory"` which could potentially confuse for users. 
This PR checks whether specified mnemonickeyfile exist or not and read it only in case it does. If there is no mnenmonickey , we create a random key and store in mnemonic.txt inside DID folder

### Testing : 
call createdid without specifying mnemonicKeyFile (or a filename that doesn't exist) 

Attaching screenshot of code execution before and after. 

Screenshot 2024-06-06 at 6 12 02 PM

### DIFF
--- a/did/did.go
+++ b/did/did.go
@@ -59,6 +59,7 @@ func InitDID(dir string, log logger.Logger, ipfs *ipfsnode.Shell) *DID {
 func (d *DID) CreateDID(didCreate *DIDCreate) (string, error) {
 	t1 := time.Now()
 	temp := uuid.New()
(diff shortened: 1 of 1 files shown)


Maneesha-rubix (approved):
Reviewed and tested

PR #175, opened 2024-06-07
issue with mnemonickey reading

harirubix:
Following #174  , there was an issue reading mnemonickeyfile . This PR fixes it.


Maneesha-rubix (approved):
Test cases:
1. mnemonic key file not passed
2. wrong mnemonic key file path
3. mnemonic key passed & did is as expected
Screenshot from 2024-06-07 14-21-25

PR #177, opened 2024-06-11
parse did to get cid digest and use this to calc lastchar

harirubix:
### Explanation:
**CID Parsing:** The cid.Decode function parses the provided CID string into a cid.Cid object.
**Extracting Multihash:** The Hash() method of the cid.Cid object extracts the underlying multihash.
**Converting to Hexadecimal:** The hex.EncodeToString function converts the binary multihash to a hexadecimal string

### DIFF
--- a/core/wallet/did.go
+++ b/core/wallet/did.go
@@ -1,5 +1,11 @@
 package wallet
 
+import (
+	"encoding/hex"
+
+	"github.com/ipfs/go-cid"
+)
+
 type DIDType struct {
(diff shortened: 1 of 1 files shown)


harirubix:
As an extension of this , we can store the CID digest in db and pick type 1 quorum based on did_digest and txnid to make it more distributed. 
Eg : We could use a XOR DHT between 2 hashes did_digest and txnid to limit quorum selection within a certain range. Open to other ideas.

ashi31 (approved):

PR #182, opened 2024-06-28
Ashita/quorum checkfor pledged token state

ashi31:
When a token is transferred, say with state t1, the quorums will check if t1 is pinned or not. If it is, means the token is being double spent. If it is not, it will pin t1 to prevent it from double spending, and after the transfer, the latest token state will become t2. So the quorums are pinning state t1, and pledging for state t2. 
Now when the token is transferred further, the new state will become t3,  and the sender will inform the old quorums, about the token state t2 being exhausted, so that the quorums can unpledge the tokens. And even if the sender doesn't send, the quorums can check the pinned state for t2 by themselves using the below APIs

To check details about all the pledged tokens and the tokenstates for which pledging is done
```
./rubixgoplatform getpledgedtokendetails -port 20001
2024-06-28T16:35:41.993+0530 [INFO]  Main: Successfully got the pledged token states info
DID	 Pledged Token	 Token State
bafybmic2jh5arz5lnwag3xfbsgtqduo3fe7ejitp3ddk7sse74a3uck5mi 
QmQqLWdiCjzqUS739NTbDFmBohc5MGoi6155JtuamSM82m  	 QmQvVDh19mTAd3YhRgbVSLPea7XU6yxJ285KWKAiD4cjBP
```
```
To check if a particular token state is exhausted ot not
./rubixgoplatform checkpinnedstate -tokenstatehash QmQvVDh19mTAd3YhRgbVSLPea7XU6yxJ285KWKAiD4cjBP -port 20001
2024-06-28T16:37:36.811+0530 [ERROR] Main: Pin not available: message="No pins available on QmQvVDh19mTAd3YhRgbVSLPea7XU6yxJ285KWKAiD4cjBP"
```

Test cases:
Same set of quorums for multiple shuttle transfers
Different sets of quorums for multiple shuttle transfers

### DIFF
--- a/core/wallet/token.go
+++ b/core/wallet/token.go
@@ -38,11 +38,12 @@ const (
 )
 
 type Token struct {
-	TokenID       string  `gorm:"column:token_id;primaryKey"`
-	ParentTokenID string  `gorm:"column:parent_token_id"`
-	TokenValue    float64 `gorm:"column:token_value"`
-	DID           string  `gorm:"column:did"`
-	TokenStatus   int     `gorm:"column:token_status;"`
+	TokenID        string  `gorm:"column:token_id;primaryKey"`
+	ParentTokenID  string  `gorm:"column:parent_token_id"`
+	TokenValue     float64 `gorm:"column:token_value"`
+	DID            string  `gorm:"column:did"`
+	TokenStatus    int     `gorm:"column:token_status;"`
+	TokenStateHash string  `gorm:"column:token_state_hash"`
 }
 
 func (w *Wallet) CreateToken(t *Token) error {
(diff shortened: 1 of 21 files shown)


core/wallet/token.go
```diff
@@ -454,6 +454,20 @@ func (w *Wallet) TokensReceived(did string, ti []contract.TokenInfo, b *block.Bl
 	return nil
 }
 
+func (w *Wallet) TokenStateHashUpdate(tokenwithtokenhash []string) {
```
harirubix, line 457: Cant you write `TokenStateHash`  to the `tokenstable` the first time when you add token to the table ? Here you are writing to db twice. First write token details & then `TokenStateHash` separately to same row.

thewebchap (changes requested):

PR #185, opened 2024-07-05
Token Chain Validation

maneeSHA-256:
Adding a CLI command as well as an API end point to validate required number of blocks of any (RBT as well as smart contract) token chain. The blocks are validated as per their type-requirement. An user can also validate all it's tokens or all it's smart contract tokens available in it's SQLite3 DB.

**Note:** Please note that, this PR is only for (RBT / smart contract) token chain validation and does not include any resolutions for failed-validations.

**TokenTransferredType:**

1. Block hash and previous block hash validation
2. Sender signature validation
3.  All quorums' signature validation

**TokenGeneratedType:**

1. Block hash validation
2. Token-owner signature validation
3. if part token, Parent token-chain latest block validation 
4. if parent token is also a part token, validate it's Parent token-chain latest block 
5. repeat step-4, until parent token is a whole token

**TokenBurntType:**

1. Block hash and previous block hash validation
2. Token-owner signature validation 

**TokenPledgedType:**

1. Block hash and previous block hash validation
2. Token-owner signature validation 

**TokenUnpledgedType:**

1. Block hash and previous block hash validation
2. Token-owner signature validation 

**TokenContractCommited:**

1. Block hash and previous block hash validation
2. Token-owner signature validation 

**TokenDeployedType:**

1. Block hash validation
2. All quorums' signature validation

**TokenExecutedType:**

1. Block hash and previous block hash validation
2. All quorums' signature validation

A few common validation checks for latest block of each token chain are:
1. Token pin check
3. Token-state pin check (in case latest block is TokenTransferredType or TokenGeneratedType )

The CLI command for token chain validation is:
`./rubixgoplatform validatetokenchain -did <user did> -token <token ID> -sctValidation -blockCount <required number of blocks to be validated> -port <port number of user node>`

- If `token` flag is provided, then  token chain of the token provided with the flag `token` will be validated
- If `token` flag is not provided, then token chain of each token in the tokens table of the user node will be validated
- `sctValidation` flag must be provided in case of smart contract token chain validation


thewebchap (approved):

arnabghose997 (changes requested):
@Maneesha-rubix 

While trying to verify a token which was sent from A to B, I am able to verify it successfully on A's end. However, on B's end, I get an error: ` DID does nto exist: did=bafybmih54d3amspwn7p4cu67wrwutmvppfkzcevabo3p6ionqzayf7lfce` . As discussed, the reason for this is that `bafybmih54d3amspwn7p4cu67wrwutmvppfkzcevabo3p6ionqzayf7lfce` is part of Server B.

My suggestion would be make the Logger statement verbose by mentioning that DIDs that tokens for a DID which are part of the server in focus, cannot be verified.

CC: @harirubix

arnabghose997 (approved):

PR #191, opened 2024-07-11
feat: periodic pledging implementation

### LINKED ISSUE #41: Migrate from PoW unpledging to Proof of Periodic Pledging
Currently Quorum nodes are running a proof of work to unpledge token to avoid token getting locked for long duration.  This can be solved by proof of periodic pledge. In this pledging mechanism , a new set of quorum is selected after a time interval (24 hours) who now are held responsible for the tokenchain's latest state. The current owner performs a self transaction to update the tokenchain state periodically and locks on previous pledges are released.

Screenshot 2023-05-11 at 10 03 00 AM

arnabghose997:
This PR intends to replace current PoW approach with Periodic Pledging. As opposed to the current approach, tokens will now be pledged for a duration of 7 days, post which token unpledging can happen. If a Trans token is in custody of owner for more than Pledge period (7 days) or if it has been transferred before the pledging period, the quorums will be able to unpledge their pledge tokens through a CLI/API call.  

Since, pledged tokens are unpledged after 7 day period. a `self-transfer` CLI/API call is provided for node to essentially perform a psuedo state change, which will invite quorums to again pledge for the trans token. 

It has been tested locally with a temporary pledged duration value of 100 seconds with both RBT transfers and Smart contract deployment and execution.

## Unpledge CLI

```
./rubixgoplatform run-unpledge -port 20005
```
It goes through all the records present in the `unpledgeQueue` table and performs the unpledging process for all transactions, if any

## Self Transfer CLI

```
./rubixgoplatform self-transfer-rbt -senderAddr bafybmidm7my4wuhqmcwwkkbrem6ytn3vrt3vpnocvjn5r3a4pyp5drc3iq -transType 2 -port 20016
```

Similar to general RBT transfer, we only pass the sender's address and the transaction type for Self Transfer.


Allen-Cherian:
###Test Results
1. Unpledge-cli : 
- When a transfer occurs, the transaction details—transaction ID, pledged token, epoch, and quorum DID—are added to the unpledgequeue table.
- Upon calling the unpledge command, the transaction details are removed from the unpledgequeue table. Concurrently, credit details are added to the credit table, and the pledged tokens are unpledged.

2. Self Transfer Cli : 
- Self transfer working fine. The transaction details is getting added in the Transaction history table and the transaction id is getting updated in the TokensTable.

Allen-Cherian (approved):

harirubix:
Working as expected . We can merge once backward compatibility is added.

arnabghose997:
A new command `unpledge-pow-pledge-tokens` has been added, as part of the migration for older nodes (`v0.0.17` and prior),  to migrate from PoW based pledging to Periodic Pledging. This command will check if there are any tokens present in the `unpledgequeue` table. If so, then the tokens will force unpledged and the `unpledgequeue` table will be dropped. The information about pledge tokens, after the Periodic Pledging upgrade, will become part of the `UnpledgeSequence` table. Following is the syntax of the command:

```
./rubixgoplatform unpledge-pow-pledge-tokens --port <quorum node's server port>
``` 

A fix has also been administered that affected old nodes migrating to Period Pledging. The format for storing quorum address in `quorummanager`  table in older versions were `<node_peer_id>.<quorum_did>` format. However, in the current implementation, we expect their format to be `<quorum_did>`. Especially while populating the `ql` attribute of `QuorumManager` struct, the values from the `address` column of  `quorummanager` table are directly appended to it. This causes issues when invoking `getPeer` method in later stages of the code.

To mitigate this, we have added a check in the `NewQuorumManager`  function, where we check if the address from `quorummanager` table is in `<node_peer_id>.<quorum_did>` format. If so, then we update it to `<quorum_did>` format before appending to `ql` attribute of `QuorumManager` struct.

thewebchap (approved):

PR #200, opened 2024-07-31
Change in allowed maximum decimal places

vaishnav-ch:
`maxDecimalPlaces` variable changed from `5` to `3` and string formatting of `GetAccountInfo()` result.

### DIFF
--- a/command/did.go
+++ b/command/did.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


arnabghose997:
The `floatPrecision` function is called, wherever we are dealing with RBT Token's value. RBT minimum value validation's check is fixed from `0.00001` to `0.001`.

harirubix (approved):

PR #207, opened 2024-08-06
Fixed the transfer of RBT tokens between DIDs that are present on the same node

### LINKED ISSUE #206: bug: Unable to transfer RBT between DIDs belonging to the same Node server
When trying to send RBT between DIDs that are present on the same server, we face an error suggesting `invalid address`. The cause of this error is that we attempt to fetch the Peer ID of the receiver DID from `DIDPeerTable`. Since, there are no records present in this table for receiver, we get the `invalid address` error.

An additional check has to kept, where we need to also check `DIDTable` of the same node where the receiver DID is created and see if the record is present or not. If present, proceed normally else throw an error.

arnabghose997:
Closes #206 and follows the suggestion mentioned in the issue.

A test case has been added which covers the transfer of RBT between DIDs present on the same server


arnabghose997:
Covered the scenario where either the Sender or Receiver is present on a Quorum node. Briefly, the fix includes adding a condition inside `updatePledgeToken` function, where we only add the token chain block when neither the sender nor receiver are present on the Quorum node. This is done to resolve a situation where addition of block to the trans token can happen twice.

Tests have been written to cover these scenarios:
- Transfer between DIDs, both present on a Non-Quorum Server
- Transfer between DIDs, both present on a Quorum Server
- Transfer between DIDS, where the Sender is present on Quorum Server and Receiver is present on Non-Quorum Server
- Transfer between DIDS, where the Sender is present on Non-Quorum Server and Receiver is present on Quorum Server

arnabghose997:
The Python tests for this PR has been run on a local MacOS system, and all tests pass successfully. Hence, the error MacOS Git workflow is associated with some Github VM issues

gklps (approved):
LGTM

PR #211, opened 2024-08-21
Secondary token

### LINKED ISSUE #98: Rubix FT
Add support for secondary tokens (fungible tokens) in rubix network.  Inspiration can be taken from part tokens module.

vaishnav-ch:
### _**Creation of FT**_

Creating a fungible token (FT) involves using the RBT. Users need to provide a name for the FT (FTName), decide how many FTs will be created (FTCount), and specify how much RBT will be used in the process (RBTAmount). Once all details are submitted, the node will create the new tokens and allow users to trade their new fungible tokens. Token value of each FT is calculated by dividing the RBT amount by the number of FTs created (Eg : If 10 FTs are created with 1 RBT, each FT Token value is 0.1).

API endpoint : **_api/create-ft_**
In this API, the data is passed as JSON input, and the keys used are:

```
“did” [String]: This is the DID of the user creating the FTs.
“ftcount” [Integer]: This represents the total number of fungible tokens to be created.
“ftname” [String]: This is the name of the fungible token being created. 
“tokencount” [Integer]: This indicates the amount of native token RBT that will be used in the creation process.
```

CLI command: **_createft_**
Flags:

```
-did: DID of the user.
-ftName: Name of the FT to be created.
-ftCount: Total number of FTs to be created.
-rbtAmount: Amount of RBTs that will be used for creation of the FT.
-port: Port number where the node is running.
```
Eg: rubixgoplatform.exe createft -ftName rubix -ftCount 10 -did bafybmieyom4y2kcngxmfy4a6ud5zx2jv5xudxeh4t4m34po2bh5eocqp6a -rbtAmount 1 -port 20050

NOTE: 
- Ensure that the FT Count and RBT amount are set such that when the RBT amount is divided by the FT Count, the result does not exceed 3 decimal places, as the network's maximum decimal precision for token values and transfers is 3. 

### **_Transfer of FT_**

FTs can be transferred from one DID to another. The user must provide details such as the sender and receiver addresses, the name and count of FTs to be transferred, and the DID of the user. The quorum consensus for FT transfers is similar to that of RBT transfers. Specifically, each quorum must hold at least 1/5 of the total token value of the FTs being transferred.

API endpoint : **_api/transfer-ft_**
In this API, the data is passed as JSON input, and the keys used are:

```
“FTCount” [Integer]: This represents the total number of FTs to be transferred.
“FTName” [String]: This is the name of the fungible token being transferred.
“comment” [String]: This field allows the sender to include a message or note regarding the transfer.
“password” [String]: This is the password required to authorize the transfer of FT.(use default “ ”).
“receiver” [String]: This is the address of the recipient who will receive the FT.
“sender” [String]: This is the address of the user initiating the transfer.
“type” [Integer]: This specifies the type of quorum to be used in the transaction being performed.
1: Quorums will be randomly picked from the network.
2: Quorums need to be specified by the sender.
```


vaishnav-ch:
squashed all commits from this branch and push new new branch `vaishnav/feature/ft`

PR #218, opened 2024-09-13
Ashita/fix/explorer url

ashi31:
While adding the explorer URL in table, it was just taking https for all URL.
Updated it to take http as well.


ashi31:
Duplicate PR

Issue #220, opened 2024-09-18
Blocker for upgrading older version nodes to latest version


Maneesha-rubix:
Before introducing BIP39 into Rubix, we had only NLSS DIDs and thus, to fetch a peer's public key we were able to initialise the peer's DID in basic mode (DID type 0) by default. But that does not work anymore after introducing BIP39, and thus we need peer's DID type, to know whether it is a NLSS DID or BIP39 DID, and initialise it's DID and fetch it's public key. 

So, to verify previous block's pledged quorum signature, the quorums need to have DID type of the previous block's pledged quorum. This situation is handled in the latest version (v0.0.18) by the sender sharing previous block quorums' peer details with the current quorums and the current quorums' peer details with the receiver. 

Currently, we have a few old nodes and old tokens, with a few RBT transfer blocks in their tokenchain, in Rubix main net which existed before integrating BIP39. Those nodes do not have peer details of the previous block quorums. Let us call the old node node_O and it's token Token_O, and it's set of quorums Qrms_O, sender in the previous block node_S and set of quorums of node_S i.e., previous block quorums Qrms_S. So the transaction in previous block is

`node_S   --- (Token_O) --->   node_O (verified and signed by Qrms_S)    (Txn in v0.0.16 or older)`

After this node_O and Qrms_O are upgraded to v0.0.18 (for now). Now, node_O wants to transfer Token_O to the receiver node_R

`node_O   --- (Token_O) --->   node_R (verified and signed by Qrms_O)    (Txn in v0.0.18 or newer)`

In v0.0.18, Qrms_O need peer details of Qrms_S to validate their signatures. But node_O does not have peer details of Qrms_S and thus is not able to share with Qrms_O, and thus Qrms_O can't validate signatures of Qrms_S in previous block, leading the Txn to fail. 

### Solution
If node_O does not have peer details of Qrms_S, it can connect to node_S and request the peer details of Qrms_S.

PR #221, opened 2024-09-23
Resolve node upgradation issue

maneeSHA-256:
This PR resolves the issue with version upgradation. It addresses the issue #220 and proposes it's solution. 

### Issue:
Before introducing BIP39 into Rubix, we had only NLSS DIDs and thus, to fetch a peer's public key we were able to initialise the peer's DID in basic mode (DID type 0) by default. But that does not work anymore after introducing BIP39, and thus we need peer's DID type, to know whether it is a NLSS DID or BIP39 DID, and initialise it's DID and fetch it's public key.

So, to verify previous block's pledged quorum signature, the quorums need to have DID type of the previous block's pledged quorum. This situation is handled in the latest version (v0.0.18) by the sender sharing previous block quorums' peer details with the current quorums and the current quorums' peer details with the receiver.

Currently, we have a few old nodes and old tokens, with a few RBT transfer blocks in their tokenchain, in Rubix main net which existed before integrating BIP39. Those nodes do not have peer details of the previous block quorums. Let us call the old node node_O and it's token Token_O, and it's set of quorums Qrms_O, sender in the previous block node_S and set of quorums of node_S i.e., previous block quorums Qrms_S. So the transaction in previous block is

node_S   --- (Token_O) --->   node_O (verified and signed by Qrms_S)    (Txn in v0.0.16 or older)

After this node_O and Qrms_O are upgraded to v0.0.18 (for now). Now, node_O wants to transfer Token_O to the receiver node_R

node_O   --- (Token_O) --->   node_R (verified and signed by Qrms_O)    (Txn in v0.0.18 or newer)

In v0.0.18, Qrms_O need peer details of Qrms_S to validate their signatures. But node_O does not have peer details of Qrms_S and thus is not able to share with Qrms_O, and thus Qrms_O can't validate signatures of Qrms_S in previous block, leading the Txn to fail.

### Solution:
If node_O does not have peer details of Qrms_S, it can connect to node_S and request the peer details of Qrms_S. 
But in older versions (< 0.0.17), senders did not have quorums' peer details in DIDPeerTable. So, if node_S can't find PeerIDs of Qrms_S in DIDPeerTable, then we can assume that the transaction is older than v0.0.17 and thus the Qrms_S must be basic mode (DID type = 0) DIDs. Now, node_S can share only the DID types of Qrms_S as 0 with node_O. node_O can then initialise DIDs of Qrms_S in basic mode and fetch their public key and verify their signatures.


gklps (approved):
LGTM

core/ping.go
```diff
@@ -258,3 +264,75 @@ func (c *Core) GetPeerInfo(p *ipfsport.Peer, peerDID string) (GetPeerInfoRespons
...
+	for i := 0; i < 5; i++ {
+		key := "did" + strconv.Itoa(i+1)
+		qrmDID := c.l.GetQuerry(req, key)
+		if qrmDID == "" {
+			continue
+		}
+		qrmDIDList = append(qrmDIDList, qrmDID)
+	}
+
+	if len(qrmDIDList) == 0 {
```
thewebchap, line 285: What is qrmDIDList is 1 - 4?  Should we make sure the prev quorumlist is always 5?

maneeSHA-256: Yeah, right. Added the changes to the latest commit.

maneeSHA-256: Yes, right. I have added the changes to the latest commit.

arnabghose997 (approved):

Issue #217, opened 2024-09-05
Return Smart Contract Hash after Smart Contract Generation API/CLI execution


arnabghose997:
We are only able to know the Smart Contract hash through Node's log, after the Smart Contract Generation API/CLI call. It should be returned after the API/CLI call

PR #226, opened 2024-09-27
Fix: Smart contract token hash not returned

Allen-Cherian:
This commit will fix the issue where the smart contract token hash is not returned. With this commit the smart contract token hash is returned as the result.  Will close the issue #217

### DIFF
--- a/command/smartContract.go
+++ b/command/smartContract.go
@@ -57,7 +57,7 @@ func (cmd *Command) generateSmartContractToken() {
 		return
 	}
 	if !basicResponse.Status {
-		cmd.log.Error("Failed to generate smart contract token", "err", err)
+		cmd.log.Error("Failed to generate smart contract token", "err", basicResponse.Message)
 		return
 	}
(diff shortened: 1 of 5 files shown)


command/smartContract.go
```diff
@@ -60,7 +60,7 @@ func (cmd *Command) generateSmartContractToken() {
 		cmd.log.Error("Failed to generate smart contract token", "err", err)
```
arnabghose997, line 60: @Allen-Cherian  While running the command `generatesct` with an invalid DID, I am getting the following error message in the terminal:

```
2024-10-14T16:50:51.121+0530 [ERROR] Main: Failed to generate smart contract token: err=<nil>
``` 

The actual error message which suggests that the input DID doesn't exists doesn't appear. This error message, however, appears on the node logs, which is expected.

In the highlighted part, replace `err` with `basicResponse.Message` to get the error message on the terminal from the command is run.

arnabghose997 (approved):
LGTM

PR #229, opened 2024-10-14
Rubix NFT Implementation

Allen-Cherian:
## Creation of NFT 

The creation of NFT API takes the file which should be converted to NFT, along with a JSON metadata and the file itself. The metadata can include any details regarding the file which is being converted to NFT. To make an NFT of multiple files, it can be compressed in a `.zip` or any other compression formats and uploaded.

API: `/api/create-nft`

In this api the data is passed as form data the keys which we use are : 

```
1. “did” [Text] - The did of the node which creates the nft
2. “metadata”[File] - JSON file which contains information about the NFT
3. ”artifact”[File] : File which is meant to be an NFT

```

Command : `create-nft`
Flags : 
-did - The did of the node
-metadata- The path of metadata file
-artifact- The path of NFT file
-port - The port of the node

## Deploy NFT 

The Deploy NFT api is similar to the smart contract deployment. While deploying an NFT, the value of NFT is assigned as 1. The amount of tokens pledged for deploying any NFT is based on this value 1. 

API : /api/deploy-nft

In this api the data is passed as json. An example of the request body is given below:

 ```
{
    "nft" :"Qmbr1isZLmSCVyi6iVJJn89pSCnwxGUFbpfHzTpyEUMKNC",
    "did" : "bafybmid23tc2ddfsmxbb5h26cyiljkylbp7fjsgpbucxlm43vzbm5p2b7y",
    "quorum_type": 2
}
```

Command : deploy-nft

Flags : 
-nft
-did
-transType
  
## ExecuteNFT
 
In this API, you can either transfer the ownership of NFT or you can simply change the internal state of the NFT by adding some  Data through self-execution. For self-execution, `receiver` param must be left empty. The `nft_value` dictates the pledge amount that Quorum needs to.

API : `/api/execute-nft`

In this api the data is passed as json. An example of the request body is given below 

```
{
    "nft":"Qmbr1isZLmSCVyi6iVJJn89pSCnwxGUFbpfHzTpyEUMKNC",
    "owner":"bafybmid23tc2ddfsmxbb5h26cyiljkylbp7fjsgpbucxlm43vzbm5p2b7y",
    "receiver":"bafybmidpuc3vcegwxjrr6xnvvwvt3txmhkjnvdgb2kbyvemilw4lptxlvm",
    "quorum_type":2,
    "comment":"Test execute nft",
    "nft_value":1.0,
    "nft_data":"Any string value user wishes to add" 
}
```

Command : execute-nft
Flags : 
-nft
-owner
-receiver
-transType
-comment
-rbtAmount
-nftData

## Subscribe NFT

For receiving an NFT a receiver node must subscribe to that particular NFT. This is similar to the subscription of smart contracts.

API : `/api/subscribe-nft`

In this api the data is passed as json. An example of the request body is given below 

```
{
    "nft":"Qmbr1isZLmSCVyi6iVJJn89pSCnwxGUFbpfHzTpyEUMKNC",
}
```

Command : subscribe-nft
Flags : 
-nft 
-port

## Get NFT Token Chain Data

This api will return the nft token chain data

API : `/api/get-nft-token-chain-data`

Query Params:

```
"nft": NFT id
"latest": If true, we only get the latest block of a NFT else we receive the entire state
```

## Dump NFT Token chain

Command: dump-nft-tokenchain

Flags:
- nft: NFT Id

### CODE REVIEW

File: server/server.go [hunk 1]
```diff
@@ -170,6 +172,9 @@ func (s *Server) RegisterRoutes() {
 	s.AddRoute(setup.APIRecoverRBT, "POST", s.AuthHandle(s.APIRecoverRBT, true, s.AuthError, false))
 	s.AddRoute(setup.APIValidateTokenChain, "GET", s.AuthHandle(s.APIValidateTokenChain, false, s.AuthError, false))
 	s.AddRoute(setup.APIValidateToken, "GET", s.AuthHandle(s.APIValidateToken, false, s.AuthError, false))
+	s.AddRoute(setup.APIGetNFTTokenChain, "POST", s.AuthHandle(s.APIGetNFTTokenChain, true, s.AuthError, false))
+	s.AddRoute(setup.APISubscribeNFT, "POST", s.AuthHandle(s.APISubscribeNFT, true, s.AuthError, false))
+	s.AddRoute(setup.APIGetNFTTokenChainData, "POST", s.AuthHandle(s.APIGetNFTTokenChainData, true, s.AuthError, false))
```
arnabghose997, line 177: This API should be `GET` based as we are only trying to retrieve NFT token chain data

setup/setup.go
```diff
@@ -77,6 +77,11 @@ const (
 	APIRecoverRBT                       string = "/api/recover-token"
 	APIValidateTokenChain               string = "/api/validate-token-chain"
 	APIValidateToken                    string = "/api/validate-token"
+	APIDeployNFT                        string = "/api/deploy-nft"
+	APITransferNFT                      string = "/api/transfer-nft"
+	APIGetNFTTokenChain                 string = "/api/get-nft-token-chain"
```
arnabghose997, line 82: A better endpoint name could be `/api/dump-nft-token-chain`. Refer this [comment](https://github.com/rubixchain/rubixgoplatform/pull/229/files#r1802492977)

core/nft.go
```diff
@@ -4,327 +4,639 @@ import (
...
+		c.log.Error("Failed to rename NFT folder", "err", err)
+		return basicResponse
+	}
+	nftTokenDetails := wallet.NFT{
+		TokenID:     nftHash,
+		DID:         nft.DID,
+		TokenStatus: 0,
+		TokenValue:  0,
+	}
+	c.w.CreateNFT(&nftTokenDetails)
```
arnabghose997, line 138: Handle the error for `c.w.CreateNFT(...)`

core/nft.go
```diff
@@ -4,327 +4,639 @@ import (
...
+		NFTFileHash:     nftFileHash,
+		NftFileInfoHash: nftFileInfoHash,
+		UserId:          userID,
+		NFTData:         nftData,
+	}
+
+	if err != nil {
+		c.log.Error("Failed to create NFT", "err", err)
+		return basicResponse
+	}
```
arnabghose997, line 106: The `err` here will always be nil. These lines must be removed

rac/rac.go
```diff
@@ -71,6 +72,12 @@ type RacPartInfo struct {
 	Value   float64
 }
 
+type RacNFTInfo struct {
+	Creator            string
+	ContentHash        string
+	ContentDescription string
+}
+
```
arnabghose997, line 80: Since the block struct for NFT is similar to that of Smart Contracts, we can remove this struct

arnabghose997 (changes requested):
@Allen-Cherian  Apart from the review comments left, following are some high level suggestions that needs to be incorporated:

1. `/api/get-nft-token-chain-data` should be a `GET` request and the same should be reflected in Swagger docs.

2. Following should be the `/api/create-nft` inputs:

```
1. did (required) : creator of NFT
2. metadata (required): a JSON file which contain all the info about the NFT
3. artifact (required) : file that represents the NFT
```
3. The casing of request params for every NFT endpoint, should follow snake_case convention, instead of camelCase or PascalCase.

4. Ensure that changes in every NFT Request params should reflect in their respective command flags as well. For instance, if the request param `NFTFileInfo` is changed to `metadata`, then the corresponding Create NFT command flag should also be changed to `metadata`.

5. `transfer-nft` should be changed to `execute-nft` and it should have two flows:

  - Ownership transfer: whenever the `receiver` field has DID in it
  - Internal State Change: NFT can undergo state change without the transfer of ownership.   The condition being that `receiver` field is empty.

Allen-Cherian:
All the requested changes have been done.
## Test Cases Tested :

1. Normal flow : A nft is created, deployed and executed. 
2. Execution from node 1 to node 2 : Works as exepected.
3. Execution from node 1 : The reciver parameter is kept as empty and nft is executed : Works as exepected.

arnabghose997 (changes requested):
@Allen-Cherian  Based on our discussions, leaving down few suggestions:

1. Dealing with scenario where Receiver and Owner DIDs are present on the same node

    When a NFT ownership transfer is initiated from current Owner to Receiver, it proceeds successfully. However, when we attempt do the reverse thereafter, the transaction fails stating that NFT information was not found in `NFTTable`
    
    For sake of understanding, lets consider first owner as `A` and second owner as `B`.
    
    What happens during first transfer (from A to B) is that the state change for the NFT in `NFTTable` happens like: `DID_A (Status: 0)  ---> DID_A (Status: 4)`. When its time for the reverse transfer, we search for the NFT owned by `DID_B`, which according to the current situation, is not present.
    
    This situation is similar to RBT transfer between DIDs on the same DID.
    
    I believe the correct state transition in the `NFTTable` should happen like this: `DID_A (Status: 0)  ---> DID_B (Status: 0)`. We can refer to the usage of [TokenTransferred](https://github.com/rubixchain/rubixgoplatform/blob/main/core/wallet/token.go#L379) in dealing with this scenario

2. Usage of IPFS Get instead of IPFS Cat operation to fetch NFT files

    IPFS Cat essentially reads the content in Bytes, which makes it difficult to identify its actual file name at source. Using IPFS Get lets directly download the file associated with the IPFS Hash.
    
    One approach is to individually call `Get` on each NFT content hash. Another approach is to create a single hash for the NFT directory, instead of individually hashing the contents of the directory
    
3. Add the key for `NFTData` (key "13", refer `block/block.go`) in the `keyMapping` variable of `command/decodeTokenChain.go` file.

arnabghose997 (approved):

PR #233, opened 2024-10-18
New explorer structure

ashi31:
Updating the explorer structures as per the new explorer format.


core/explorer.go
```diff
@@ -28,10 +37,11 @@ type ExplorerClient struct {
 }
 
 type ExplorerDID struct {
-	PeerID    string `json:"peerid"`
+	PeerID    string `json:"peer_id"`
 	DID       string `json:"user_did"`
-	IPAddress string `json:"ipaddress"`
+	IPAddress string `json:"ip_addess"`
```
gklps, line 42: IPAddress is not needed

core/explorer.go
```diff
@@ -40,18 +50,32 @@ type ExplorerMapDID struct {
...
-	DeployerDID string   `json:"deployer_did"`
-	ExecutorDID string   `json:"executor_did"`
-	//BlockHash   string   `json:"block_hash"`
+type Token struct {
+	TokenID    string  `json:"token_id"`
+	TokenValue float64 `json:"token_value"`
+}
+
+type ChildToken struct {
+	TokenID    string  `json:"child_token_hash"`
```
gklps, line 59: can you update the TokenID as ChildTokenID

core/explorer.go
```diff
@@ -84,6 +152,12 @@ func (c *Core) InitRubixExplorer() error {
 		return err
 	}
 
+	err = c.s.Init(ExplorerUserDetailsTable, &ExplorerUser{}, true)
+	if err != nil {
+		c.log.Error("Failed to initialise storage ExplorerUserDetails ", "err", err)
+		return err
+	}
+
 	url := "deamon-explorer.azurewebsites.net"
```
gklps, line 161: shouldnt these URL be coming from explorer db?

core/explorer.go
ashi31: Yes, removed it. When it was still there, I was getting error that its mandatory field

core/explorer.go
ashi31: Done, is TokenValue fine? Also, the json string is different in the latest swagger. So udpated that as well.

core/explorer.go
ashi31: It is coming from db when making requests. But while initialising the ExplorerURLTable, its adding the mainnet or testnet explorer link. This was there since earlier. Didn't make any changes here. Should I remove these links?

core/explorer.go
```diff
@@ -188,12 +238,14 @@ func (c *Core) InitRubixExplorer() error {
 		return err
 	}
 	url := "rexplorer.azurewebsites.net"
+	if c.testNet {
```
gklps, line 241: The explorer URL should be coming from DB right?

core/explorer.go
```diff
@@ -696,20 +833,25 @@ func (c *Core) ExpireUserAPIKey() {
...
-		}
-		eu.APIKey = ""
-		err = c.s.Update(ExplorerUserDetailsTable, &eu, "did=?", eu.DID)
-		if err != nil {
-			c.log.Error(err.Error())
-		}
-		c.log.Info(fmt.Sprintf("%v for DID %v", er.Message, eu.DID))
+		go func(eu ExplorerUser) {
+			var er ExplorerResponse
+			err = c.ec.SendExploerJSONRequest("POST", ExplorerExpireUserKeyAPI, &eu, &er)
```
gklps, line 838: SendExploerJSONRequest - update the function spelling

gklps (commented):
How many parallel API is being sent from the node? Can we test the max limit possible without any blocker, if already done pls share the result

PR #235, opened 2024-10-23
fix: re-added block to notify previous quorums to remove trans tokens from `TokenStateHash` table that have their token state exhausted

### LINKED ISSUE #234: fix: re-adding the code block that informs old quorums to update records in `TokenStateHash` for exhausted Token state hashes
After merging of periodic pledging PR #191 and due to some rebasing thereafter, a particular code block (Refer [here](https://github.com/rubixchain/rubixgoplatform/blob/82763c105490f5355b343fec61440d83694dbefb/core/quorum_initiator.go#L565-L591)) got removed, and hence it isn't present in `v0.0.18`.

This code block informs the old Quorums to remove the records for those Trans Tokens in `TokenStateHash` table, whose states are exhausted.

In the `v0.0.18`, the absence of above code block doesn't affect the RBT transfer. It only affects the time it will take for the Pledge tokens to get unpledged.

For instance, say we transfer 2 RBT tokens from A to B. Then, B decides to transfer these tokens to C, before the Pledge period of 7 days. The expected flow is that B will inform old quorums to remove the records, associated with these trans tokens, from their `TokenStateHash` table. But, currently that won't happen. Hence, when the `/run-unpledge`  API is ran, the associated will not unpledge (which it is supposed to). Instead, they will unpledge after the 7 day pledge period is over.

arnabghose997:
Closes #234

### DIFF
--- a/core/quorum_initiator.go
+++ b/core/quorum_initiator.go
@@ -567,6 +567,35 @@ func (c *Core) initiateConsensus(cr *ConensusRequest, sc *contract.Contract, dc 
 			c.log.Error("Pledge finlaity not achieved", "err", err)
 			return nil, nil, pledgeFinalityError
 		}
+
+		//Checking prev block details (i.e. the latest block before transferring) by sender. Sender will connect with old quorums, and update about the exhausted token state hashes to quorums for them to unpledge their tokens.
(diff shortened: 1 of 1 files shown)


gklps (approved):
Lgtm

PR #236, opened 2024-10-23
Resolving errors in Fetch smart contract 

KryptSai:
In this PR, I changed the smart contract API from a POST request to a GET request due to errors related to the Content-Type, which was set to multipart/form-data. These errors occurred when passing the required data through Swagger while calling the API.
This PR also resolves the smart contract folder renaming error that occurred when fetching the smart contract from the command line as well as from the API, in cases where the node already contains the smart contract folder.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


core/core.go
```diff
@@ -427,6 +427,21 @@ func (c *Core) RenameSCFolder(tempFolderPath string, smartContractName string) (
 	}
 	return scFolderName, err
 }
+func (c *Core) Foldercheck(smartContractName string) interface{} {
```
arnabghose997, line 430: Rename the function to `IsSmartContractDirectoryPresent` and change the return type to `bool`

core/core.go
```diff
@@ -427,6 +427,21 @@ func (c *Core) RenameSCFolder(tempFolderPath string, smartContractName string) (
...
+    folderName := c.cfg.DirPath + "SmartContract/" + smartContractName
+    info, err := os.Stat(folderName)
+    if err != nil {
+        c.log.Error("Unable to get the file info, error is: ", err)
```
arnabghose997, line 434: No need for the Error logger to be added here, since the necessary logs will be present outside of this function

core/core.go
```diff
@@ -427,6 +427,21 @@ func (c *Core) RenameSCFolder(tempFolderPath string, smartContractName string) (
...
+        return false
+    }
+
+    // Return the folder name if it's a directory, otherwise return false
+    if info.IsDir() {
+        return folderName
+    }
```
arnabghose997, line 441: Remove this codeblock since the function is only expected to return Boolean and not string

core/core.go
```diff
@@ -427,6 +427,21 @@ func (c *Core) RenameSCFolder(tempFolderPath string, smartContractName string) (
...
+
+    return false
```
arnabghose997, line 443: The default return here must be `true`

arnabghose997 (changes requested):
Added some review comments for the function responsible for checking if the Smart Contract directory is present or not

arnabghose997 (approved):

PR #238, opened 2024-11-01
Rubix FT

vaishnav-ch:
**FT**

**_Creation of FT_**

Creating a fungible token (FT) involves using the RBT. Users need to provide a name for the token (FTName), decide how many tokens to be created (FTCount), and specify how much RBT will be used in the creation process (Tokencount/rbtAmount). Once all details are submitted, the node will create the new tokens and allow users to trade their new fungible tokens. Token value of each FT is calculated by dividing the RBT amount by the number of FTs created (eg: If 10 FTs are created with 1 RBT, each FT Token value is 0.1). 

API endpoint : api/create-ft
In this API, the data is passed as JSON input, and the keys used are:
“did” [String]: This is the Decentralized Identifier of the user creating the FTs. It uniquely identifies the user on the network.
“ftcount” [Integer]: This represents the total number of fungible tokens to be created.
“ftname” [String]: This is the name of the fungible token being created. 
“tokencount” [Integer]: This indicates the amount of native token RBT that will be used in the creation process. It is essential for verifying that sufficient funds are available for creating the FTs.
Eg:

{
  "did": "bafybmieyom4y2kcngxmfy4a6ud5zx2jv5xudxeh4t4m34po2bh5eocqp6a",
  "ftcount": 10,
  "ftname": "rubix",
  "tokencount": 1
}

CLI command: create-ft
Flags:
-did: DID of the user.
-ftName: Name of the FT to be created.
-ftCount: Total number of FTs to be created.
-rbtAmount: Amount of RBTs that will be used for creation of the FT.
-port: Port number where the node is running.
Eg: rubixgoplatform.exe create-ft -ftName rubix -ftCount 10 -did bafybmieyom4y2kcngxmfy4a6ud5zx2jv5xudxeh4t4m34po2bh5eocqp6a -rbtAmount 1 -port 20050

NOTE: 
- Ensure that the FT Count and RBT amount are set such that when the RBT amount is divided by the FT Count, the result does not exceed 3 decimal places, as the network's maximum decimal precision for token values and transfers is 3. 

**_Transfer of FT_**

FTs can be transferred from one DID to another. The user must provide details such as the sender and receiver addresses, the name and count of FTs to be transferred, and the DID of the user. The quorum consensus for FT transfers is similar to that of RBT transfers. Specifically, each quorum must hold at least 1/5 of the total token value of the FTs being transferred.

API endpoint : /api/initiate-ft-tranfer
In this API, the data is passed as JSON input, and the keys used are:
“FTCount” [Integer]: This represents the total number of FTs to be transferred.
“FTName” [String]: This is the name of the fungible token being transferred.
“comment” [String]: This field allows the sender to include a message or note regarding the transfer.
“password” [String]: This is the password required to authorize the transfer of FT.(use default “ ”).
“receiver” [String]: This is the address of the recipient who will receive the FT.
“sender” [String]: This is the address of the user initiating the transfer.
“type” [Integer]: This specifies the type of quorum to be used in the transaction being performed.
1: Quorums will be randomly picked from the network.
2: Quorums need to be specified by the sender.
“creatorDID” [String]: This field defines the creator of an FT. 

Eg:

{
  "FTCount": 0,
  "FTName": "string",
  "comment": "string",
  "creatorDID": "string",
  "password": "string",
  "receiver": "string",
  "sender": "string",
  "type": 0
}

CLI Command: transfer-ft 
Flags: 
-ftName [String]: The name of the FT being transferred.
-ftCount [Integer]: The total number of FTs to be transferred.
-senderAddr [String]: The address of the user sending the tokens.
-receiverAddr [String]: The address of the recipient who will receive the FTs.
-comment [String]: An optional message or note regarding the transfer.
-fp [String]: The password required to authorize the transfer of FT. (Default “”)
-transType [Integer]: This specifies the type of transaction being performed.
-port [String]: The port number where the node is running.
-creatorDID [string]: DID of the creator of the FT.

Eg: ./rubixgoplatform transfer-ft -ftName rubix -ftCount 1 -senderAddr bafybmifzhiucxmq7uwm67iksbm52esb6yfzxtq6gpktlouvzzx3fn36eya -receiverAddr bafybmieyom4y2kcngxmfy4a6ud5zx2jv5xudxeh4t4m34po2bh5eocqp6a -password “” -transType 2 -creatorDID bafybmia4s7sjljsrwvfz2mnqyk7garf744d5tjbwrofnhtmcu7erghirvu -port 20051

NOTE: 
-	-fp is used when there is a password already created during the creation of DID.
-	-creatorDID flag is used when there are multiple FTs with the same FTName.

**_Get FT Info by DID_**

There may be multiple FTs associated with the same DID. The 'Get FT Info by DID' function will provide a list of these FTs along with their count. Users only need to supply the DID and the port number.

API Endpoint : api/get-ft-info-by-did
In this API, the data is passed as query input and the key used is :

“did” [String]: DID of the user.

Eg:

CLI Command: get-ft-info-by-did

Flags:
-did [String] : DID of the User.
-port [String]: The port number where the node is running.

Eg: ./rubixgoplatform get-ft-info-by-did -did bafybmifzhiucxmq7uwm67iksbm52esb6yfzxtq6gpktlouvzzx3fn36eya  -port 20001

**_Get FT token chain data_**

The token chain of a fungible token (FT) begins with the genesis block, which is crucial for establishing the foundational parameters of the token. This block contains essential information regarding the creation of the FT, including the RBT token IDs that serve as parent token IDs, the value assigned to the FT, and the timestamp of its creation. Following the genesis block, every transaction involving the FT is recorded as a new block in the token chain, which includes details such as quorum signatures and other pledging information. This structure ensures that all transactions are securely documented and can be referenced later, allowing users to fetch the entire token chain for verification and historical context.

API Endpoint: api/get-ft-token-chain-data

In this API, the data is passed as query input and the key used is :

“tokenID” [String]: TokenID of the FT


command/command.go
```diff
@@ -206,6 +214,10 @@ var commandsHelp = []string{"To get tool version",
 	"This command will initiate a self RBT transfer",
 	"This command will unpledge all the pledged tokens",
 	"This command will unpledge all PoW based pledge tokens and drop the unpledgequeue table",
+	"This command will create FT",
```
gklps, line 217: Orders are mismatched. Confirm with -help command if the description are coming as expected

command/ft.go
```diff
@@ -0,0 +1,135 @@
...
+		cmd.log.Error("FT Name can't be empty")
+		return
+	}
+	br, err := cmd.c.CreateFT(cmd.did, cmd.ftName, cmd.ftCount, cmd.rbtAmount)
+	if strings.Contains(fmt.Sprint(err), "no records found") || strings.Contains(br.Message, "no records found") {
+		cmd.log.Error("Failed to create FT, No RBT available to create FT")
+		return
+	}
+	if err != nil {
+		cmd.log.Error("Failed to create FT", "err", err)
```
gklps, line 32: Can we have a proper error log here. Why would the FT creation fail here and 2 lines below

command/ft.go
```diff
@@ -0,0 +1,135 @@
...
+	if cmd.receiverAddr == "" {
+		cmd.log.Info("Receiver address cannot be empty")
+		fmt.Print("Enter Receiver DID : ")
+		_, err := fmt.Scan(&cmd.receiverAddr)
+		if err != nil {
+			cmd.log.Error("Failed to get Receiver DID")
+			return
+		}
+	}
+	isAlphanumericSender := regexp.MustCompile(`^[a-zA-Z0-9]*$`).MatchString(cmd.senderAddr)
```
gklps, line 66: What validation are you doing here

core/ft.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+	channel := c.GetWebReq(reqID)
+	if channel == nil {
+		c.log.Error("Failed to get did channels")
+		return
+	}
+	channel.OutChan <- &br
+}
+
+func (c *Core) createFTs(reqID string, FTName string, numFTs int, numWholeTokens float64, did string) error {
```
gklps, line 39: any reason for float64 to have variable name as numWholeTokens

core/ft.go
```diff
@@ -0,0 +1,559 @@
...
+
+	if len(FT) != 0 {
+		c.log.Error("FT Name already exists")
+		return fmt.Errorf("FT Name already exists")
+	}
+	if dc == nil {
+		return fmt.Errorf("DID crypto is not initialized")
+	}
+
+	// Validate input parameters
```
gklps, line 57: Can we make this a switch case than having a lot of if else conditions

core/ft.go
```diff
@@ -0,0 +1,559 @@
...
+	for _, token := range wholeTokens {
+		parentTokenIDsArray = append(parentTokenIDsArray, token.TokenID)
+	}
+	parentTokenIDs := strings.Join(parentTokenIDsArray, ",")
+	for i := 0; i < numFTs; i++ {
+		racType := &rac.RacType{
+			Type:        c.RACFTType(),
+			DID:         did,
+			TokenNumber: uint64(i),
+			TotalSupply: 1,
```
gklps, line 98: ?

core/quorum_initiator.go
```diff
@@ -636,6 +644,163 @@ func (c *Core) initiateConsensus(cr *ConensusRequest, sc *contract.Contract, dc
...
+				qrmInfo.DIDType = &qDidType
+			}
+			//add quorum details to the data to be shared
+			qrmInfo.DID = qdid
+			qrmInfo.PeerID = qpid
+			sr.QuorumInfo = append(sr.QuorumInfo, qrmInfo)
+		}
+
+		var br model.BasicResponse
+		err = rp.SendJSONRequest("POST", APISendReceiverToken, nil, &sr, &br, true)
```
gklps, line 701: Why is FT acting as a RBT trnx?

core/model/ft.go
```diff
@@ -0,0 +1,30 @@
...
+	DID        string  `json:"did"`
+	FTName     string  `json:"ftname"`
+	FTCount    int     `json:"ftcount"`
+	TokenCount float64 `json:"tokencount"`
+}
+
+type TransferFTReq struct {
+	Receiver   string `json:"receiver"`
+	Sender     string `json:"sender"`
+	FTName     string `json:"FTName"`
```
Allen-Cherian, line 13: It is better to follow the json naming convention as ft_name and ft_count

core/model/ft.go
```diff
@@ -0,0 +1,30 @@
...
+	FTCount    int    `json:"FTCount"`
+	Comment    string `json:"comment"`
+	Type       int    `json:"type"`
```
Allen-Cherian, line 16: Instead of using type it will be better to use quorum_type because, since type is a reserved word, it can create issue while we write contract and all.

core/quorum_initiator.go
```diff
@@ -34,6 +34,7 @@ const (
 	SelfTransferMode
 	PinningServiceMode
 	NFTExecuteMode
+	FTTrasnferMode
```
Allen-Cherian, line 37: A minor typo here

setup/setup.go
```diff
@@ -76,12 +76,17 @@ const (
 	APIInitiatePinRBT                   string = "/api/initiate-pin-token"
 	APIRecoverRBT                       string = "/api/recover-token"
 	APIValidateTokenChain               string = "/api/validate-token-chain"
+	APICreateFT                         string = "/api/create-ft"
+	APIDumpFTTokenChainBlock            string = "/api/dump-ft-token-chain"
+	APIInitiateFTTransfer               string = "/api/initiate-ft-tranfer"
```
Allen-Cherian, line 81: In transfer the 's' is missing

core/quorum_recv.go
```diff
@@ -908,11 +912,15 @@ func (c *Core) updateReceiverToken(
...
-
-	updatedTokenStateHashes, err := c.w.TokensReceived(receiverDID, tokenInfo, b, senderPeerId, receiverPeerId, pinningServiceMode, c.ipfs)
+	var FT wallet.FTToken
+	FT.FTName = ftinfo.FTName
+	updatedTokenStateHashes, err := c.w.TokensReceived(receiverDID, tokenInfo, b, senderPeerId, receiverPeerId, pinningServiceMode, c.ipfs, FT)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to update token status, error: %v", err)
 	}
+	if FT != (wallet.FTToken{}) {
+		c.updateFTTable(receiverDID)
```
arnabghose997, line 922: Please handle the error for this function and return the error

core/wallet/token.go
```diff
@@ -729,20 +729,103 @@ func (w *Wallet) TokensReceived(did string, ti []contract.TokenInfo, b *block.Bl
...
+			err = w.s.Write(FTTokenStorage, &FTInfo)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// Update token status and pin tokens
+		tokenStatus := TokenIsFree
+		role := OwnerRole
+		ownerdid := did
+		if pinningServiceMode {
```
Allen-Cherian, line 799: PinningService function still here

command/command.go
vaishnav-ch: Order updated. Confirmed with -h flag.

command/ft.go
vaishnav-ch: Error handled and log updated.

core/quorum_recv.go
vaishnav-ch: Error handled.

core/wallet/token.go
vaishnav-ch: Removed unused parts.

core/ft.go
vaishnav-ch: Changed to int.

core/quorum_initiator.go
vaishnav-ch: Separate functions added for FT consensus.

core/model/ft.go
vaishnav-ch: changed to `ft_name and ft_count`.

core/model/ft.go
vaishnav-ch: Type changed to quorumType.

core/ft.go
```diff
@@ -0,0 +1,550 @@
...
+		FTOwner := blk.GetOwner()
+		ft := &newFTs[i]
+		ft.CreatorDID = FTOwner
+		err = c.w.CreateFT(ft)
+		if err != nil {
+			c.log.Error("Failed to create fractional token", "err", err)
+			return err
+		}
+	}
+	c.updateFTTable(did)
```
arnabghose997, line 255: Please handle the error

vaishnav-ch: Error handled.

Allen-Cherian (approved):

arnabghose997 (approved):

core/quorum_initiator.go
```diff
@@ -334,6 +348,11 @@ func (c *Core) initiateConsensus(cr *ConensusRequest, sc *contract.Contract, dc
 		reqPledgeTokens = 1
 	case SmartContractExecuteMode, NFTExecuteMode:
 		reqPledgeTokens = sc.GetTotalRBTs()
+	case FTTransferMode:
```
thewebchap, line 351: What is the difference between this and reqPledgeTokens in  `case RBTTransferMode, NFTSaleContractMode, SelfTransferMode, PinningServiceMode`: ?

core/quorum_recv.go
```diff
@@ -658,6 +658,123 @@ func (c *Core) quorumNFTConsensus(req *ensweb.Request, did string, qdc didcrypto
 	return c.l.RenderJSON(req, &consensusReply, http.StatusOK)
 }
 
+func (c *Core) quorumFTConsensus(req *ensweb.Request, did string, qdc didcrypto.DIDCrypto, cr *ConensusRequest) *ensweb.Result {
```
thewebchap, line 661: What specific operations are you doing here when compared to `quorumRBTConsensus` method?

thewebchap (changes requested):

command/ft.go
```diff
@@ -0,0 +1,132 @@
...
+	if !isAlphanumericSender || !isAlphanumericReceiver {
+		cmd.log.Error("Invalid sender or receiver address. Please provide valid DID")
+		return
+	}
+	if !strings.HasPrefix(cmd.senderAddr, "bafybmi") || len(cmd.senderAddr) != 59 || !strings.HasPrefix(cmd.receiverAddr, "bafybmi") || len(cmd.receiverAddr) != 59 {
+		cmd.log.Error("Invalid sender or receiver DID")
+		return
+	}
+
+	transferFtReq := model.TransferFTReq{
```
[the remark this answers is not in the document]
vaishnav-ch: Yes. Will add similar validation for creatorDID also. `ftcount` validation is happening at initiateFT() in core/ft.go, can be added here also, for transType too.

core/quorum_initiator.go
vaishnav-ch: obtaining `reqPledgeTokens` is similar to RBTTransferMode, NFTSaleContractMode, SelfTransferMode, PinningServiceMode. Just added as a case, no difference. Can be added with `RBTTransferMode` case also.

core/quorum_recv.go
vaishnav-ch: Pinning service checks is not in FT consensus. No extra operation is doing in `quorumFTConsensus` other than from `quorumRBTConsensus`.

core/quorum_initiator.go
thewebchap: This switch case scope is to only to obtain `reqPledgeTokens` .

thewebchap: Also , majority of code of FT and RBT (part RBT) are same. I would prefer a generalised function to handle both . And we do minimal token based operations within.

vaishnav-ch: getting the total token value to be transferred is same for FT transfer and RBT transfer. Also the steps involved in the FT transfer is very same to RBT transfer.

command/ft.go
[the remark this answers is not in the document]
thewebchap: Sure , we should have both client side and server (node) side validations.

thewebchap (approved):

PR #255, opened 2024-12-03
User can define the address

vaishnav-ch:
`NodeAddress` field was hard-coded to "localhost". Now users can define the address using `-addr` flag.


arnabghose997 (approved):

thewebchap (approved):

gklps:
We need to update the same in updateconfig also

gklps (approved):

PR #273, opened 2024-12-26
fix for FT Transfer between DIDs on the same node

arnabghose997:
This PR aims to fix a scenario where FT is transferred between two DIDs present on the same node. Following are the changes made:

- In `FTTokensTransffered` function, a check has been added which skip the entire function for the scenario where Sender DID and Receiver DID are present on same node. This avoids duplicated addition of token blocks in the same node.

- Before initiating consensus, we check if the RecieverDID is part of the same node as SenderDID. If so, then we assign the peer for ReceiverDID as `c.peerID`.


Allen-Cherian (approved):

gklps (approved):
LGTM

PR #288, opened 2025-01-16
updating parameters as per swagger

ashi31:
Updating the explorer API parameters as per the swagger.

### DIFF
--- a/core/ft.go
+++ b/core/ft.go
@@ -553,7 +553,7 @@ func (c *Core) initiateFTTransfer(reqID string, req *model.TransferFTReq) *model
 
 	eTrans := &ExplorerFTTrans{
 		FTBlockHash:     AllTokens,
-		CreatorDID:      []string{creatorDID},
+		CreatorDID:      creatorDID,
 		SenderDID:       did,
 		ReceiverDID:     rdid,
 		FTName:          req.FTName,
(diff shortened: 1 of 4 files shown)


core/ft.go
```diff
@@ -562,7 +562,7 @@ func (c *Core) initiateFTTransfer(reqID string, req *model.TransferFTReq) *model
 		FTSymbol:        "TODO",
```
gklps, line 562: is "TODO" an entry to explorer?

core/ft.go
```diff
@@ -562,7 +562,7 @@ func (c *Core) initiateFTTransfer(reqID string, req *model.TransferFTReq) *model
...
 		Comments:        req.Comment,
 		TransactionID:   td.TransactionID,
-		PledgeInfo:      PledgeInfo{PledgeDetails: pds.PledgedTokens, PledgedTokenList: pds.TokenList},
+		PledgeInfo:      PledgeInfoNFT{PledgeDetails: pds.PledgedTokens, PledgedTokenList: ConvertToNFTTokens(pds.TokenList)},
 		QuorumList:      extractQuorumDID(cr.QuorumList),
 		Amount:          FTsForTxn[0].TokenValue * float64(req.FTCount),
```
gklps, line 567: Can we ensure this is rounded to 3 decimal points?

gklps (changes requested):
Having "TODO" as a value shared to explorer isn't a good practise. Either send the value for that for now or avoid that key

core/ft.go
ashi31: Updated it to N/A

core/ft.go
ashi31: done

gklps:
@ashi31 backend has been updated with the request body you proposed. Ensure those changes

PR #290, opened 2025-01-17
Return DID for existing public key

maneeSHA-256:
When a DID for a particular public key already exists in the ode's database, the `/api/request-did-for-pubkey` API was throwing error: "UNIQUE constraint failed: DIDTable.did". This PR is returning the existing DID for the requested public key instead of throwing an error.


Gokul-Sabu:
LGTM

Gokul-Sabu (approved):

PR #296, opened 2025-02-19
Add nft value and nft data to /api/deploy-nft

Allen-Cherian:
## Changes Introduced  
1. **Enhanced `/api/deploy-nft` Endpoint**  
   - Added `nftValue` and `nftData` fields to the API request/response. 

2. **Updated `/api/get-nft-token-chain-data` Endpoint**  
   - Included `epoch` in the response.

3. **Removed Committed Tokens List from NFT Genesis Block**  
   - The committed tokens list, which was previously added to the genesis block of an NFT, has been **removed**.  
   - **Reason:** Unlike smart contracts, NFTs do not require token commitments at deployment. Keeping this list was unnecessary and inconsistent with the NFT deployment process.
  
4. **Add the NFTData to the genesis block while deploying NFT**

### DIFF
--- a/block/block.go
+++ b/block/block.go
@@ -747,8 +747,8 @@ func (b *Block) GetChildTokens() []string {
 	return util.GetStringSliceFromMap(b.bm, TCChildTokensKey)
 }
 
-func (b *Block) GetEpoch() int64 {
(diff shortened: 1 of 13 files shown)


arnabghose997 (approved):
LGTM

PR #298, opened 2025-02-27
Fix for GetTxnDetailsByID function

### LINKED ISSUE #297: ignore the err for GetTransactionByDID when startDateStr != ""  or  endDateStr!= ""
https://github.com/rubixchain/rubixgoplatform/blob/f97220bf71d6daad3854cd0bddbeeacb4d7b012e/core/txndetails.go#L29

didn't check the err, when startDateStr != ""  or  endDateStr!= "" ,  
as a result, may return an empty result when err.

ashi31:
Updated the function for GetTxnDetailsByDID. Checking the filters for date range and role properly.


gklps (approved):
LGTM

PR #305, opened 2025-03-12
User create api update -  New NFT/FT Struct

ashi31:
Updating the User create API, and warning message.
If a user is already created, generates the API Key


gklps:
Do we need #288  ?

PR #309, opened 2025-03-24
Fixing Node Upgradation for Migrated Tokens

maneeSHA-256:
### Issue : 
The genesis block of migrated tokens is signed by some Advisory node. Advisory nodes are fixed set of nodes with one DID each.


gklps (approved):
lgtm

PR #312, opened 2025-04-11
Error handling incase of Tokenchain Sync issue

Allen-Cherian:
This PR addresses two key issues :

1. **Tokenchain Sync Issue**:  
   Previously, if the `GetAllTokenBlocks` function failed to find a specified `blockID`, an error was returned stating that the tokenchain block does not exist, and the `tokenStatus` was set to 12. The improvement in this PR ensures that in such cases, all nodes with a pin on the token are retrieved, and their roles for that pin are checked. This additional information provides better clarity on the root cause of the issue.

2. **Unpledging Tokens Issue**:  
   Earlier, pledged tokens were not unpledged properly, resulting in an error: "tokens should be in pledged state before unpledging." This was caused by the SQLite database not being updated correctly. This PR resolves the issue by ensuring proper database updates during the unpledging process.


gklps (approved):
lgtm

PR #316, opened 2025-04-15
Token Chain Pruning

maneeSHA-256:
### Description
This PR is making the transaction flow smoother by syncing only the genesis and latest blocks (and the current transaction block) of the trans tokens and their parent tokens while transacting, and syncing the rest of the blocks as a background process, without interfering the transaction flow.

### Fail Safe
In case the background syncing fails, sync status of that particular token is updated as _SyncIncomplete_. Then whenever the node restarts, all the tokens with _SyncIncomplete_ sync status, are processed to sync all the unsynced blocks in a background go-routine. Once the syncing is completed, token status is updated to _SyncCompleted_

### Tested Scenarios in TestNet 
Set 1 :
Quorum set : QS1 (q1 - q7)
sender : S1
receiver : R1

Set 2 : 
Quorum set : QS2 (q11 - q15) 
sender : S2
reciver : R2

1. Transfer 0.32 RBT (say token T1) from S1 -> R1 with quorums QS1
2. Transfer same 0.32 RBT (T1) from R1 -> S1 with quorums QS1
3. Repeat step-1 and step-2 2 more times with same token T1
4. Transfer 0.29 RBT (child token of T1, say T2) from S1 -> S2 with quorums QS2
5. Transfer 0.29 RBT (T2) from S2 -> R2 with quorums QS2
6. Checked sync status of T1 and T2 for each node
7. Dumped token chain of T1 for all and compared : all have block height - 8
8. Dumped token chain of T2 for  all except R1 and compared : S1 and QS1 have block height - 2 and S2, R2 and QS2 have block height - 3


gklps (approved):
lgtm

Allen-Cherian (approved):

PR #325, opened 2025-05-14
Changes associated with NFT, Smart Contracts and other minor fixes 

arnabghose997:
Smart Contracts:

- Added the following request parameters during API call being made to callback URL, which avoids the need for recalling Rubix Endpoints from the Contract Linker functions:
  - `smart_contract_data`: Input being passed to the WASM Contract
  - `initiator_did`: DID which initiated the Execute Smart Contract API transaction
 
- Added `initiatorSignature`, `signedMessage` and `initiatorDid` in the `/api/get-smart-contract-token-chain-data` API's reponse

- PeerID has been added in the Smart Contract IPFS Info enabling the Smart Contract fetch operation to sync the token chain as well

NFT:

- Added `transactionID` in the `/api/get-nft-token-chain-data` API's reponse

- Fixed NFT self-execution by allowing non-owner DIDs to execute the NFT which earlier was failing

- Creation of NFT will only involve adding the files to IPFS. Therefore, the action to add new NFT record's into DB has been shifted under Deploy NFT API, making it independed of Create NFT API. @Allen-Cherian  This needs to be replicated for Smart Contract as well.

Other Fixes:

- Under `ChainDB` struct, changed the embedded struct to reference type (from `level.DB` to `*level.DB`). Earlier, the SSTable were not generated because having `level.DB` caused issues with levelDB internal Mutex's state.  This change solves the issue allowing bulk data ( > 12000 records) to be created without any panic issues

- Loggers added in the Rubix Daemon to show peer IDs of Bootstrap nodes that are connected and display errors upon failed attempt to connect.

### DIFF
--- a/command/bootstrap.go
+++ b/command/bootstrap.go
@@ -10,12 +10,12 @@ func (cmd *Command) addBootStrap() {
 		cmd.log.Error("Peers required for bootstrap. Use flag -peers to provide peers separated by a ','")
 		return
 	}
-	for _, peer := range cmd.peers {
-		if !strings.HasSuffix(peer, "/") {
-			cmd.log.Error(fmt.Sprintf("Invalid bootstrap peer : %s", peer))
-			return
-		}
-	}
+	// for _, peer := range cmd.peers {
+	// 	if !strings.HasSuffix(peer, "/") {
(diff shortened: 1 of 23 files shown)


Allen-Cherian (approved):
LGTM

PR #326, opened 2025-05-14
Add Rust Code Analysis: Infinite Loop Detection

Allen-Cherian:
This PR introduces Rust code analysis during contract generation and fetching. If the Rust code contains an infinite loop, contract generation will be blocked. The same analysis is also performed when fetching the contract.

## Core Logic
- Analyzes specific loop types (`while`, `for`, `loop`) and recursive function calls for conditions that may lead to infinite loops.

## Recursive Traversal
- Processes nodes recursively to check the entire syntax tree.

### While Loops
- Flags loops with conditions always `true` or unmodified variables.

### Rust's `loop {}` Syntax
- Flags unless a `break` statement is present to exit the loop.

### For Loops
- Detects non-terminating iterators (e.g., `std::iter::repeat`) and checks for `break` statements.

### Recursive Function Calls
- Flags direct recursion as potential infinite recursion.

### Complex While Conditions
- Dynamic conditions (dependent on external states or computations) may not be fully analyzed.

### External Break Conditions
- Break conditions dependent on external threads or asynchronous contexts may be missed.

### Custom Iterators
- Does not analyze termination for user-defined iterators.

### Indirect Recursion
- Only detects direct recursion, not circular dependencies between functions.

### Unreachable Code
- Flags loops in unreachable code (e.g., within `if false`) unnecessarily.


core/smart_contract.go
```diff
@@ -218,19 +217,18 @@ func (c *Core) FetchSmartContract(requestID string, fetchSmartContractRequest *F
 	binaryCodeFileDestPath := filepath.Join(binaryCodeFilePath, "binaryCodeFile")
 
 	// Read the content of binaryCodeFile
-	binaryCodeContent, err := ioutil.ReadAll(binaryCodeFile)
+	binaryCodeContent, err := io.ReadAll(binaryCodeFile)
 	if err != nil {
 		c.log.Error("Failed to read binary code file", "err", err)
 		return basicResponse
 	}
```
arnabghose997: Set `basicResponse.Message` string and replicate for other `if err` blocks that follows

core/smartcontract_analyser.go
```diff
@@ -0,0 +1,302 @@
...
+func (c *Core) AnalyseCode(code []byte) bool {
+	c.log.Info("Analyzing code for potential infinite loops and recursion")
+	// Initialize parser
+	parser := sitter.NewParser()
+	parser.SetLanguage(rust.GetLanguage())
+
+	// Parse the code
+	tree, err := parser.ParseCtx(context.Background(), nil, code)
+	if err != nil {
+		c.log.Error((fmt.Sprintf("Error parsing code: %v\n", err)))
```
arnabghose997: Remove the extra parentheses pair here

core/smartcontract_analyser.go
```diff
@@ -0,0 +1,302 @@
+package core
```
arnabghose997, line 1: Few pointers:

- The functions present here would be standalone instead of being methods of `Core` struct.
- Instead of logging, the messages should be returned from the function

arnabghose997 (changes requested):

core/smartcontract_analyser.go
Allen-Cherian: All the methods apart from the AnalyseCode have been removed from the ```Core``` struct

core/smartcontract_analyser.go
```diff
@@ -17,62 +16,58 @@ type FunctionInfo struct {
 	CallNodes []*sitter.Node
 }
 
-func (c *Core) AnalyseCode(code []byte) bool {
-	c.log.Info("Analyzing code for potential infinite loops and recursion")
+func (c *Core) AnalyseCode(code []byte) error {
```
arnabghose997: Keep it as a standalone function, and make the function private

arnabghose997: Based on discussion with @Allen-Cherian , since `server` package requires this function, it would remain a public function while not being a method of `Core` struct

arnabghose997 (approved):
LGTM

PR #334, opened 2025-06-03
FT flag updation

vaishnav-ch:
This PR will add a flag `-ftNumStartIndex` which can used during creation of FT. This flag is used to define a specific FT number when creating same FT on different nodes or DIDs. 
Eg:

- DID A of Node A created 1000 FTs named **TEST** (FT Number will be 0 to 999)
- DID B of Node A or DID X of Node X can create FT named **TEST** starting from FT number 1000 by defining flag `-ftNumStartIndex` as 1000.


gklps (approved):
LGTM

PR #337, opened 2025-06-04
fixing getalldid issue with DIDs created with CreateDidFromPubKey

maneeSHA-256:
For DIDs created with the api : /api/request-did-for-piubkey, the did_dir column in DIDTable was not being updated as "root". Due to which DID was not treated as native DID while fetching all DIDs of the node using API : /api/getalldid. This PR is resolving this issue by properly enabling verift=ying the native DIDs and updating the did_dir column of those DIDs.


gklps (approved):
LGTM

PR #407, opened 2025-10-16
Add file validation for smart contract generation and remove unused schema

Allen-Cherian:
This PR introduces validation checks for files uploaded during smart contract generation to ensure they are of the correct type. It also includes a minor cleanup by removing an unused parameter from the generation process.

### Changes

-   **Added WASM File Validation:**
    -   The binary file uploaded for the smart contract must now have a `.wasm` extension.
    -   If a non-`.wasm` file is provided, the process will fail with the error: `Binary code file must be a .wasm file`.

-   **Added Rust Source File Validation:**
    -   The raw source file  must now have a `.rs` extension.

-   **Code Cleanup:**
    -   Removed the `schemaFile` parameter from the smart contract generation function.
    -   **Reason:** This parameter was unused throughout the codebase, and its removal simplifies the function.

### DIFF
--- a/client/smart_contract.go
+++ b/client/smart_contract.go
@@ -10,7 +10,6 @@ import (
 type SmartContractRequest struct {
 	BinaryCode string
 	RawCode    string
-	SchemaCode string
 	DID        string
 	SCPath     string
 }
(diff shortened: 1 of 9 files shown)


arnabghose997 (approved):
LGTM

PR #425, opened 2025-11-28
Fix: Same node transfer 

Allen-Cherian:
The issue was : When we transfer token count greater than 50, in the same node between 2 different dids, the transfer was failing. The reason being the transaction id and tokenstatus as pending was not getting updated properly in the tokenstorage table. Now the table is getting updated properly.


gklps (approved):
LGTM

PR #431, opened 2025-12-31
checks at FullNode side

KryptSai:
In this PR,  we are adding the following validations at fullnode side while syncing them from other nodes.
1. While adding an incoming block to fullnode's levelDB it checks whether,  previous block_ID of the incoming block matches with block_ID of the existing latest block or not. 
ex: let's say FullNode is having 10blocks, 11th block it is syncing from other node, then 11th block's previous block ID should be same as existing 10th block's ID.
2. If the incoming block is a transfer block, FullNode checks whether the sender of the incoming block matches with the owner of the latest existing block or not.
ex: let's say FullNode is having 10blocks, 11th block it is syncing from other node, then 11th block's senderDID, should be same as existing 10th block's OwnerDID.

**Fullnode handles syncing token  in the following 3 ways:** 
Lets say FullNode has 7 blocks of a particular token t1.
1. case:If the publisher node publishes any blocks with block numbers  greater than 7th(lets say from 8th to 10th)
In this case, FullNode need to sync the incoming block. So while adding each block to Fullnode's levelDB it checks the above conditions. It will add only the above checks pass. 
2. case: If the publisher node publishes a block with block number 7.
In this case, Fullnode checks whether is it the same block which is coming or is it the different block. If it is the different block, then it will add it into a double spend tokens table with DID1 &DID2 having the same block.
3. case: If the publisher node publishes a block with block number less than 7(lets say 5).
In this case, Fullnode doesn't need to sync. So it checks whether the incoming block(5th block), is it the same as the 5th block which the FullNode is having, If the Fullnode has a different 5th block than what publisher is publishing, It will add the token to double spend tokens table with DID1 & DID2 as the owners.

More possible exploits but not resolving for now(If we assume  all the publisher nodes are under our control, no need to resolve this case):
case description:
let's say nodeA has 8 blocks.  A has done double spend transaction, A -> B and A -> B' both with blocknumber 9. Now 
B->C,  C->D , D->E transactions has happend. So currently E is having 12th block. B' is having 9th block only. When Firstly, if FullNode sync from B', FullNode will get till 9th block. After that if E also publishes tokenchain with length 12. Fullnode checks, previous blockID of the incoming 10th block won't match with the already existing 9th block. so we will note down both DIDs as the same token holder for later reference while giving the tokens to them. If we do in this way any node can simply add a dummy block to their levelDB and publish with a different DID as a publisherDID then Fullnode treat it as a double spend transaction and note down both DIDs to distribute the tokens later. 

Some more checks to discuss: 
1. lets say a Fullnode has received part token's genesis block,  then fullnode should get it's parent tokenID and check whether it has already burnt block of the parent token with it, but the issue would be both,  parent tokens burnt block and part tokens genesis block,  will be get created at almost same point of time, so there is a chance that fullnode might not received in the same order in which they both get published, so this check might not be a good fit for the fullnodes.


command/diagnostic.go
```diff
@@ -148,10 +148,13 @@ func (cmd *Command) dumpTokenChain() {
 		}
 	}
 
-	elems := strings.Split(cmd.token, "_")
-	if len(elems) < 2 {
-		cmd.log.Error(fmt.Sprintf("Invalid token: %v", cmd.token))
-		return
+	if cmd.assetType == "rbt" || cmd.assetType == "RBT" {
```
arnabghose997: Lets' have only "rbt" as value to denote RBT Asset type

core/wallet/token.go
```diff
@@ -1717,6 +1712,30 @@ func (w *Wallet) ReadFailedToSyncTokensFromTable(tokenID string) (*model.FailedT
...
+		parentTokenID,
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// If no child tokens exist, return nil (as requested)
+	if len(childTokens) == 0 {
+		return nil, nil
```
arnabghose997, line 1733: Two scenarios:

- If this condition is expected to return an error, the error should be returned instead of `nil`
- If this condition is not expected to return an error, instead of returning `nil` for list, an empty list should be returned

core/fullnode.go
```diff
@@ -285,102 +329,110 @@ func (c *Core) processRegularTransfer(newEvent *model.PubSubTxnInfo, txnBlock *b
...
+					info := &model.FailedToSyncTokenDetailsInfo{
+						TokenID:   tokenId,
+						TokenType: tokenType,
+						AssetType: newEvent.AssetType,
+						Did:       newEvent.PublisherDID,
+						Reason:    fmt.Sprintf("failed to sync chain,err: %v", err),
+					}
+
+					if err := c.w.AddFailedTokensToTable(info); err != nil {
+						c.log.Error("Failed to record failed token sync in DB", "token", tokenId, "error", err)
```
arnabghose997, line 356: Please return error here

core/fullnode.go
```diff
@@ -285,102 +329,110 @@ func (c *Core) processRegularTransfer(newEvent *model.PubSubTxnInfo, txnBlock *b
 				}
 				err = c.SyncFullTokenChainForFullNode(p, *tokenSyncInfo)
 				if err != nil {
-					c.log.Error("failed to sync token chain for token ", tokenId, "error", err)
-					return fmt.Errorf("failed to get latest block for token %s - may need sync", tokenId)
-
+					handled, _ := c.HandleSyncErrorAsDoubleSpent(err, tokenId, newEvent.AssetType, tokenType, newEvent.PublisherDID, previousOwner, txnBlockOwner, fmt.Sprintf("%s, %s both dids are claiming the same token or same did would have double spent the token,dual ownership issue", previousOwner, txnBlockOwner))
```
arnabghose997: Return the error here

core/fullnode.go
```diff
@@ -285,102 +329,110 @@ func (c *Core) processRegularTransfer(newEvent *model.PubSubTxnInfo, txnBlock *b
...
-					ClaimedOwnerII: newEvent.PublisherDID,
-					ErrorMessage:   "publisher DID mismatch with current owner in burnt block",
-				}
-				err = c.StoreDoubleSpentTokenInfo(doubleSpentTokenInfo)
+				// return fmt.Errorf("mismatch of blocks detected: latest block number=%d, current block number=%d", latestBlockNumber, currentBlockNumber)
+			} else {
+				incomingBlockID := txnBlockID
+				publishersLatestBlockNumber := currentBlockNumber
+				//get Fullnode block whose block number is publisher side latest blockID
+				fullnodesideBlockBytes, err := c.w.GetFullNodeTokenBlockByNumber(tokenId, tokenType, publishersLatestBlockNumber)
```
arnabghose997, line 370: The `err` is not returned here. Please return it

core/fullnode.go
```diff
@@ -285,102 +329,110 @@ func (c *Core) processRegularTransfer(newEvent *model.PubSubTxnInfo, txnBlock *b
...
-		// Validate receiver for transfers
-		if receiverDid != "" {
-			if currentOwner != receiverDid {
-				return fmt.Errorf("receiver DID mismatch: expected %s, got %s", receiverDid, currentOwner)
+			receivedBlock.GenesisBlock = txnBlock
+		}
+		// Do all 3 checks, If any check fails, handle the failure here also just before adding the block to leveldb.
+		err = c.ValidateIncomingTokenBlock(*txnBlock, latestTokenBlock, tokenId, p, newEvent.AssetType)
+		if err != nil {
+			handled, _ := c.HandleSyncErrorAsDoubleSpent(err, tokenId, newEvent.AssetType, tokenType, newEvent.PublisherDID, previousOwner, txnBlockOwner, fmt.Sprintf("Incoming block %s is wrong,error %v", txnBlockID, err))
```
arnabghose997: error is not returned here

core/fullnode.go
```diff
@@ -457,20 +519,123 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
...
+			//connect to publisher and fetch complete token chain
+
+			tokenSyncInfo := &TokenSyncInfo{
+				TokenID:   tokenId,
+				TokenType: tokenType,
+				AssetType: newEvent.AssetType,
+			}
+			err = c.SyncFullTokenChainForFullNode(p, *tokenSyncInfo)
+			if err != nil {
+				handled, _ := c.HandleSyncErrorAsDoubleSpent(err, tokenId, newEvent.AssetType, tokenType, newEvent.PublisherDID, previousOwner, txnBlockOwner, fmt.Sprintf("%s, %s both dids are claiming the same token or same did would have double spent the token,dual ownership issue", previousOwner, txnBlockOwner))
```
arnabghose997: error is not returned here

core/fullnode.go
```diff
@@ -457,20 +519,123 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
...
+				} else {
+					c.log.Info("Recorded failed token sync in DB", "token", tokenId)
+					return nil
+				}
+			}
+			// check if token exists in postgres table, add if doesn't
+			err := c.ReadTokenContentFromPSQL(tokenId, newEvent.AssetType)
+			if err != nil {
+				if err := c.AddTokenContentToPSQL(tokenId, newEvent.AssetType); err != nil {
+					c.log.Error("failed to add token's ipfs content to psql db, err: %v", err)
```
arnabghose997, line 619: Please return the error

core/fullnode.go
```diff
@@ -457,20 +519,123 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
...
+				info := &model.FailedToSyncTokenDetailsInfo{
+					TokenID:   tokenId,
+					TokenType: tokenType,
+					AssetType: newEvent.AssetType,
+					Did:       newEvent.PublisherDID,
+					Reason:    fmt.Sprintf("failed to sync chain,err: %v", err),
+				}
+
+				if err := c.w.AddFailedTokensToTable(info); err != nil {
+					c.log.Error("Failed to record failed token sync in DB", "token", tokenId, "error", err)
```
arnabghose997, line 609: Please return the error here

core/fullnode.go
```diff
@@ -457,20 +519,123 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
...
+					AssetType:      newEvent.AssetType,
+					TokenType:      tokenType,
+					PublisherDID:   newEvent.PublisherDID,
+					ClaimedOwnerI:  fullnodesideBlockOwner,
+					ClaimedOwnerII: txnBlockOwner,
+					ErrorMessage:   fmt.Sprintf("%s, %s both dids are claiming the same token,dual ownership issue", fullnodesideBlockOwner, txnBlockOwner),
+				}
+				// store double spent token info in DoubleSpentTokens table, and remove it from respective tokens table
+				err = c.StoreDoubleSpentTokenInfo(doubleSpentTokenInfo)
+				if err != nil {
```
arnabghose997, line 651: error is not returned here

core/fullnode.go
```diff
@@ -479,6 +644,14 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
 		}
 	}
 
+	// Do all 3 checks, If any check fails, handle the failure here also just before adding the block to leveldb.
+	err = c.ValidateIncomingTokenBlock(*txnBlock, latestTokenBlock, tokenId, p, newEvent.AssetType)
+	if err != nil {
+		handled, _ := c.HandleSyncErrorAsDoubleSpent(err, tokenId, newEvent.AssetType, tokenType, newEvent.PublisherDID, previousOwner, txnBlockOwner, fmt.Sprintf("Incoming block %s is wrong,error %v", txnBlockID, err))
```
arnabghose997: error is not returned here

core/fullnode.go
```diff
@@ -494,10 +667,23 @@ func (c *Core) processContractExecution(newEvent *model.PubSubTxnInfo, txnBlock
...
+
 	if err := c.AddTokenToRespectiveTable(tokenId, currentOwner, receivedBlock, newEvent, syncStatus); err != nil {
 		return fmt.Errorf("failed to add contract token to table: %v", err)
 	}
 
+	// check if token exists in postgres table, add if doesn't
+	err = c.ReadTokenContentFromPSQL(tokenId, newEvent.AssetType)
+	if err != nil {
+		if err := c.AddTokenContentToPSQL(tokenId, newEvent.AssetType); err != nil {
+			c.log.Error("failed to add token's ipfs content to psql db, err: %v", err)
```
arnabghose997, line 712: error is not returned here

arnabghose997 (changes requested):
Ensure the errors are returned properly

PR #433, opened 2026-01-05
return transaction id along with response while deploying and executing smart contracts

Allen-Cherian:
return transaction id along with response while deploying and executing smart contracts


gklps (approved):
LGTM

PR #439, opened 2026-01-12
Fix Higher RBT Fetch Issue; Async Consensus & Unpledge Issue

maneeSHA-256:
## Problems

1. **Higher RBT Fetch for Transfer** : The Optimized RBT fetching algorithm was sometimes fetching more RBT than requested, leading to transaction failure.
2. **Asynchronous Conensus for RBT transfer** : When RBTs are transferred in a loop, it can lead to transaction failure if next transfer is initiated before completion of the current transfer. 
3. **Asynchronous Unpledge Request** : Previous quorums are requested to initiate unpledge process after the sender sends the trans-tokens to receiver. If the unpledge request fails, then the transaction fails at sender end, and the sender updates the token status to be free. But the receiver has already received the tokens and have updated the tokens metadata in DB. This leads to double spend.

## Solutions 

1. **Higher RBT Fetch for Transfer** : Select `1` token less than requested amount, and fetch the remaining amount using the un-optimized RBT fetch algorithm.
2. **Asynchronous Consensus for RBT transfer** : Remove asynchronous consensus and return transaction completion status only after transaction is complete (with success/ failure status).
3. **Asynchronous Unpledge Request** : Send the unpledge request to previous quorums asynchronously, so that the transaction does not fail even when the unpledge request fails.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -34,7 +34,7 @@ const (
 )
 
 const (
-	version string = "0.1_G"
+	version string = "0.1_mig"
 )
 const (
 	VersionCmd                     string = "-v"
--- a/command/did.go
+++ b/command/did.go
@@ -260,7 +260,7 @@ func (cmd *Command) SignatureResponse(br *model.BasicResponse, timeout ...time.D
 		if br.Result == nil {
 			return br.Message, true
 		}
-		
+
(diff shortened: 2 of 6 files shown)


gklps (approved):
LGTM

PR #440, opened 2026-01-15
Fix Remove Stale DID API 

maneeSHA-256:
### Problem 
The API `/api/remove-stale-did` checks RBT balance and FT balance before removing the DID from the network. But this is creating issue in Xell migration transfers, as we are not planning to transfer existing FTs in OLD DIDs to the new DID within the Xell migration.

### Solution 
Do not check for FT balance while removing stale DID. RBT balance check is sufficient.


gklps (approved):
LGTM

PR #446, opened 2026-02-04
Add missing keys in all blocks (epoch key and token value key)

vaishnav-ch:
Add the **epoch** and **token value** in the blocks where it was missing

Epoch added in:
- FT's genesis block
- RBT's burnt block for FT
- Token pinned as service block
- Smart contract committed tokens block
- Test token genesis block
- Faucet token genesis block
- Unpledge block

Token Value added in:
- Smart contract deploy block
- Smart contract execute block
- Token pinned as service block
- Transfer block
- Smart contract committed tokens block
- Pledge block
- Unpledge block (token value added by reading tokens table)

Reference ID added in:
- Pledge block (Transaction ID is passed in the field for navigating transfer block from pledged block)


gklps (approved):
LGTM

PR #447, opened 2026-02-04
Remove RAC related functions

vaishnav-ch:
RAC was used during the creation of any asset in the network. This is changed to not use RAC struct for creating any asset.
RAC remove from:
- FT creation
- Part token creation
- Local test token creation
- Faucet test token creation

Removed all functions(including helpers) related to RAC block creation & initialization.


gklps (approved):
LGTM

PR #450, opened 2026-02-04
remove SmartContract and NFT field from TokenChainBlock

Allen-Cherian:
This PR :
1. Removes the SmartContract and NFT fields from TokenChainBlock.  
2. Change the NFTData and SmartContractData to Data. A point to note in this case is that, we have not edited the TransInfo structure. There the SmartContractData and NFTData fields still exists. Another point to be noted is that the now we are getting all the information which we used to get from the SmartContract field from the TransInfo struct which was already part of the tokenchainBlock.


core/quorum_recv.go
maneeSHA-256, line 1:
```
td := &model.TransactionDetails{
			Amount:          b.GetTokenValue(),
}
```
This Amount is NOT TokenValue stored in block. We have to calculate the sum of all tokens from the structure of the tokenIDs, and assign the sum to td.Amount

maneeSHA-256 (dismissed):
everything else LGTM

core/quorum_recv.go
Allen-Cherian: We can make that addition once we push the new token_level:token_number structure

maneeSHA-256 (approved):
LGTM

PR #461, opened 2026-02-08
added IPFS.Add() before major IPFS cat, unpin, get and tokenID DHT check operations

arnabghose997:
This PR adds IPFS Add operation to convert the TokenID to TokenHash wherever there is any major IPFS operations such as IPFS cat, unpin, get and tokenID

**Please Note**

`pinCheck()` function has been modified where a code block to convert TokenID to TokenHash has added inside. Hence, moving forward it should be ensured that only **Token ID** is passed to `pinCheck()`, not Token Hash


gklps (approved):
lgtm

gklps (approved):
lgtm

maneeSHA-256 (approved):

PR #466, opened 2026-02-16
In generate test RBT, making token_level and token_number exactly similar to Faucet tokens or Mainnet tokens. 

KryptSai:
While generating test net tokens, lets say If user generates more tokens than in the current existing level(according to the token Map). It will go to next level. Example: In level-0, user can create 56 tokens, after 56 tokens, token_level will be 1, token number will start from 0.


Allen-Cherian (approved):
LGTM

PR #468, opened 2026-02-17
fixes associated with FT transfer

arnabghose997:
This PR intends to fix the issues that occurs during the transfer of FT tokens between DIDs present on different nodes.

In the earlier case, the FT tokens while being created were not being added to IPFS because of the new token structure. This caused the issue on the receiver's end, when it attempted to unpin the tokenHash, which in reality never existed in the swarm network in the first place. Hence, a IPFS Add call was inserted under `(*Core).createFTs` function to address this issue,

In addition to this, a validation has been added under all the IPFS related methods (except for `Add(...)`), to check if the input is a IPFS CID (v0 or v1). If it isn't, then the input is added to IPFS. This has been done to avoid the need for looking at places where IPFS related function calls are being made, and adding an `Add` operation before it.

### DIFF
--- a/core/ft.go
+++ b/core/ft.go
@@ -1,9 +1,9 @@
 package core
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
-	"bytes"
 	"log"
 	"math"
 	"regexp"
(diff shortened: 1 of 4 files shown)


Allen-Cherian (approved):
LGTM

Allen-Cherian (approved):
LGTM

Issue #465, opened 2026-02-13
Need to add more Validations at the quorum side


KryptSai:
**Case description:** 
Let's say  DID1 is having a genesis block of a token(t) and transferred it to DID2 with quorum(Q), Now if the same DID1 transfers the same token (by making the token status to Free and without deleting the 1st transfer block) transfer happens,  even though quorum does the token state hash check, because quorum checks pin on the token statehash of tokenID-1st blockID, which is not yet exhausted so check will pass and sender will be able to send the token to DID3. Issue is quorum is not checking whether incoming transblock's senderDID is same as the existing latest block's ownerDID or not? so that's why when a DID1 intiates double spending transaction without deleting 1st transaction block, it will go through. 
**Following checks need to add at quorum side:**
1. For every incoming transaction block's senderDID is same as the existing latest block's ownerDID or not?
2. If transaction token is a part token with genesis block, quorum should check, whether its parent token is already burnt or not? 
3. Is this transaction block actually forming a blockchain or not? Check if previous block hash stored in the incoming transaction block matches with the latest block hash of each trans-token involved.

PR #471, opened 2026-02-19
block Owner and previous blockID checks at quorum side

KryptSai:
In this PR, we have added 1st , 3rd  checks discussed in github issue #465. This will prevent double spending of a token from A->B, and  A->C without deleting the 1st transaction A->B block.


Allen-Cherian (approved):
LGTM

Allen-Cherian (approved):

PR #481, opened 2026-03-02
fix for not updating the local token level in LocalTestTokenInfo table.

KryptSai:
This PR is to fix the issue in updating local test token info table.


Allen-Cherian (approved):
LGTM

PR #615, opened 2026-03-23
Update initiateConsensus function to incorporate pledging logic 

Allen-Cherian:


core/transaction.go
```diff
@@ -25,7 +26,7 @@ func (c *Core) initiateTransaction(reqID string, request *models.TransactionRequ
 	resp := &model.BasicResponse{
 		Status: false,
 	}
-	ctx := context.TODO()
+	ctx := context.Background()
```
arnabghose997: Use `c.Ctx` instead

Allen-Cherian: updated with c.Ctx

util/helpers.go
```diff
@@ -45,3 +47,16 @@ func ParseAddress(addr string) (string, string, bool) {
 	//TODO:: Validation
 	return peerID, did, true
 }
+
+func HasNFTOwnershipTransfer(nfts []models.NFTInfo) bool {
+	for _, nft := range nfts {
+		if nft.NFTId == "" {
```
arnabghose997: This is not needed

Allen-Cherian: Removed that function entirely and NFT sync logic has been updated

arnabghose997 (approved):

PR #645, opened 2026-04-02
added methods in post consensus persistence to increase and decrease token denom; fixed transaction syncing by making DB operations atomic

arnabghose997:


arnabghose997:
Sync operation handled in #655

PR #646, opened 2026-04-02
fixed the issue of transaction value being zero at quorum side in the…

KryptSai:
In this PR, we have fixed an issue where transaction value which is passed to quorum in pledge request is zero, so quorum was not pledging anything.


arnabghose997:
Fixed in #647

PR #725, opened 2026-06-24
updating the network check, checks whether network mode which is ther…

KryptSai:
In this check, we are updating the network check while validating transaction Info fields, This check ensures that when quorum or fullnode gets a transaction to validate, it will check whether the network mode in the transaction is same as the network mode which is there in its config.toml file. Previously it used to check that whether network mode which is there in the transaction Info is one among the mainnet, testnet or localnet or not.

Test cases:
1. while quorum is running in testnet, initiator initiated a transaction with network mode as mainnet, quorum rejected the transaction.
2. while fullnode is running in testnet, initiator has published a transaction with network mode as mainnet, fullnode rejected the transaction.

### DIFF
--- a/core/consensus/checks.go
+++ b/core/consensus/checks.go
@@ -45,7 +45,7 @@ func validateDID(did string, label string) error {
 	return nil
 }
 
-func ValidateTransactionInfoFields(txnInfo *models.TransactionInfo) error {
+func ValidateTransactionInfoFields(txnInfo *models.TransactionInfo, configuredNetworkMode string) error {
 	if err := validateDID(txnInfo.Initiator, "initiator"); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


core/consensus/checks.go
```diff
@@ -77,6 +91,29 @@ func ValidateTransactionInfoFields(txnInfo *models.TransactionInfo) error {
 	return nil
 }
 
+func configuredNetworkModeFromFlags(testnet, mainnet, localnet bool) (string, error) {
```
Allen-Cherian: This function duplicates util.GetNetworkMode, which performs the same operation. Please remove it and call util.GetNetworkMode here instead.

Allen-Cherian (approved):
LGTM

PR #728, opened 2026-07-08
fix(did): keep DID ownership and peerID consistent across register/create

Allen-Cherian:
# Summary

This PR fixes a bug where DID ownership information could become stale across nodes. Specifically, a node re-registering the same DID with a new libp2p peerID would not propagate that change because existing DID records were never updated. It also fixes inconsistencies in how local DID ownership (`local` flag) was maintained.

# Problem

When a node executes `registerdid`, it publishes a signed `PeerMap` (DID + peerID) on the `rubix_did` pubsub topic. Receiving nodes process this message in `peerCallback` and persist the DID → peerID mapping in the `dids` table.

However, writes were guarded by an existence check:

```go
if exists, _ := c.w.IsDIDExists(didInfo.DID); !exists {
    c.w.CreateOrUpdateDID(didInfo)
}
```

As a result, announcements for DIDs that already existed in the database were silently ignored—even if they contained a different peerID.

This caused problems when a node restarted with a new libp2p identity (common in Docker deployments where peerIDs are not persistent). Other nodes continued using the stale peerID, making the DID unreachable.

The failure was silent because the skipped path produced no log output, making it difficult to distinguish from a pubsub delivery failure.

This guard was introduced in commit `416296f9` during the DID/quorum refactor. It replaced an unconditional `AddDIDPeerMap(...)` with a `CreateOrUpdateDID(...)` call wrapped in `!exists`. While likely intended to avoid redundant writes, it also prevented legitimate ownership updates from being applied.

A related issue affected the `local` ownership flag:

- If a node created a DID that already existed as `local=false` (from an earlier remote announcement), `CreateDID` returned early and never promoted the row to `local=true`.
- If ownership moved to another node, the previous owner incorrectly retained `local=true` because only the `peer_id` was updated.

# Solution

Introduce `wallet.UpsertDIDPeer`, a dedicated upsert used only by `peerCallback`:

```sql
INSERT INTO dids (did, peer_id, local, algo_id)
VALUES ($1, $2, $3, $4)
ON CONFLICT (did) DO UPDATE
SET
    peer_id = EXCLUDED.peer_id,
    local   = EXCLUDED.local
WHERE dids.peer_id IS DISTINCT FROM EXCLUDED.peer_id
RETURNING true;
```

This provides the following behavior:

- **New DID:** Inserts a new row.
- **Existing DID with a different peerID:** Updates `peer_id` and `local` while preserving `did` and `algo_id`.
- **Existing DID with the same peerID:** Treated as a no-op.

`peerCallback` always passes `local=false`, and it already ignores announcements originating from its own peerID. Therefore, receiving a verified announcement with a different peerID correctly indicates that another node now owns the DID.

The update is safe because `peerCallback` verifies the announcement signature before writing. Only the holder of the DID's private key can successfully advertise ownership.

`peerCallback` now performs a single database upsert instead of separate existence checks and updates, reducing the operation from three database round-trips to one. Logging has also been updated to emit messages only when a row is actually inserted or updated.

Additionally, the early `IsDIDExist` return has been removed from `CreateDID`. Local DID creation now always goes through `CreateOrUpdateDID`, allowing an existing `local=false` record to be promoted to `local=true` and ensuring local ownership is recorded correctly.

# Compatibility

`CreateOrUpdateDID` itself is unchanged. Its existing callers—including the synchronization paths in `sync.go` and `transaction_chain.go` that rely on preserving `peer_id` when an empty value is supplied—continue to behave exactly as before.

# Changes

- **core/wallet/did.go**
  - Add `UpsertDIDPeer` to atomically insert or update DID ownership (`peer_id` + `local`).

- **core/peer.go**
  - Replace the `IsDIDExists` + `GetPeerID` + update sequence with `UpsertDIDPeer`.
  - Log only when a DID is actually inserted or updated.

- **core/did.go**
  - Remove the `IsDIDExist` early return from `CreateDID` so local ownership is always recorded.

### DIFF
--- a/core/did.go
+++ b/core/did.go
@@ -121,6 +121,17 @@ func (c *Core) RegisterDID(reqID string, did string) {
 }
 
 func (c *Core) registerDID(reqID string, did string) error {
+	didInfo, err := c.w.GetDID(did)
+	if err != nil {
+		c.log.Error("failed to fetch DID from dids table, err", err)
+		return fmt.Errorf("failed to fetch DID from dids table, err: %v", err)
+	}
+	// check if the did is local to the node,
+	// do not allow registerdid if not local
+	if !didInfo.Local || didInfo.PeerID != c.peerID{
+		c.log.Error("DID is not local, cannot register")
+		return fmt.Errorf("DID is not local, cannot register")
+	}
 	dc, err := c.SetupDID(reqID, did)
 	if err != nil {
 		return fmt.Errorf("DID is not exist")
--- a/core/did.go
+++ b/core/did.go
@@ -36,10 +36,10 @@ func (c *Core) CreateDID(didCreate *types.DIDCreate) (did string, err error) {
 		}
 	}
 
-	if c.IsDIDExist(did) {
-		return did, nil
-	}
-
+	// Always record local ownership, even if the DID row already exists. A prior
+	// remote rubix_did announcement may have inserted it as local=false; creating
+	// it here means this node holds the mnemonic, so upsert it as local=true.
+	// CreateOrUpdateDID's ON CONFLICT DO UPDATE promotes the existing row.
 	dt := &models.DID{
 		DID:    did,
 		PeerID: c.peerID,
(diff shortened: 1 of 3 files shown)


gklps (approved):
LGTM

PR #729, opened 2026-07-21
Fullnode Tokenchain Sync Issue

maneeSHA-256:
### Background

Fullnodes were intermittently rejecting legitimate transactions with:
```
TokenChainIntigrityCheck: token <X> chain mismatch after sync from <peer>: local latest <A> != expected <B>
```

even though the transaction itself was valid. Root-caused via targeted debug logging and a live reproduction: when an RBT transfer requires splitting a token, the wire protocol emits **two separate transactions** — a split/genesis transaction and the transfer transaction that spends its output. Because the fullnode's worker pool processes transactions concurrently with no per-token locking, `IsParentTokenBurnt` and `ValidateMinterAllowlist` — which needed a peer's chain data purely to verify a fact about a token's genesis — were fetching the peer's **full current chain with no exclusion list**, and durably persisting whatever came back via the normal chain-sync/apply path. If a sibling transaction had already been committed by the peer but not yet validated by this fullnode, that sibling's entry got silently written into `fullnode_tokenchain`, corrupting the chain-tip state the sibling's own (correctly-scoped) validation was still relying on.

### Changes

**1. Genesis-only peer fetch API, used by `IsParentTokenBurnt` / `ValidateMinterAllowlist`**
Added `/rubix/v1/internal/fetch_genesis_transaction` (`core/sync.go`, `core/core.go`, route registered in `core/transaction.go`) — returns *only* the token's genesis (position 0) transaction from the peer's local DB, and — unlike `SyncTransactionChainsFromPeer` — never writes anything to local storage. `IsParentTokenBurnt`'s fallback lookup and `ValidateMinterAllowlist`'s whole-token minter resolution now use this instead of the general chain-sync-and-persist path, closing the write side-effect that caused the corruption. Wired through `ValidateTransaction` → `fullnode.go`'s `processSingleTransaction` and `quorum_initiator.go`'s `initiateConsensusHandler`.

**2. Self-referential shortcut in `IsParentTokenBurnt`**
When the token being checked has an empty `PreviousTransactionID`, its genesis *is* the transaction currently being validated — the answer (`CommittedTokens`) is already in memory as `currentTxnInfo`, so no DB lookup or peer call is needed at all. This eliminates the network round-trip entirely for the common case (a split transaction validating its own freshly-minted children), which was also the routine trigger for the race in practice.

**3. Backfill `PreviousTransactionID` for split-derived pledge tokens (`ReqPledgeToken`)**
`CollectRBTTokens` builds a freshly-split token's `TokenInfo` before it knows the split's own transaction ID, leaving `PreviousTransactionID` empty even when the token isn't actually a genesis token relative to the pledge/consensus transaction. `transaction_builder.go` already had this fixup for the sender's own transfer tokens; `ReqPledgeToken` (pledge token path) did not. Left unfixed, this fed a false "empty = self-referential" signal into fix #2. above, causing `IsParentTokenBurnt` to check the wrong transaction's `CommittedTokens` and reject legitimate pledges — in both the quorum's own pre-signing validation (`initiateConsensusHandler`) and the fullnode's post-consensus validation.

**4. Resolve pledge tokens' genesis from the pledging quorum, not the initiator (`ValidateMinterAllowlist`)**
A pledge token's chain history lives with the DID that pledged it (the quorum), not the transaction initiator. `ValidateMinterAllowlist` was always fetching from `txnInfo.Initiator` regardless of token type; it now tracks each token's correct owning DID (mirrors the same `syncFromPeerDID` pattern `TokenChainIntegrityCheck` already used) so pledge-token genesis lookups target the right peer.

### Testing

- **Root-cause confirmation**: added temporary debug logging across the sender, quorum, and fullnode paths (DB reads/writes in `core/wallet`, sync requests/responses in `core/sync.go`, publish events in `types/pubsub.go`, split detection in `core/parts`) and reproduced the failure live twice against real nodes (recover → split → transfer), capturing full timestamped logs that traced the corrupting write to the exact `IsParentTokenBurnt` call and peer response.
- **Static verification**: `go build ./...`, `go vet ./...`, and `go test ./...` all pass clean after each change; updated the two `checks_test.go` call sites for the new `ValidateTransaction`/`IsParentTokenBurnt` signatures.
- **Unit tests**: existing `core/consensus` suite (`TestValidateTransaction_*`, etc.) passes unchanged, confirming no regression to existing validation behavior for non-split, non-pledge transactions.


Allen-Cherian (approved):
LGTM

gklps (approved):
LGTM

PR #739, opened 2026-09-02
Fix Denom Array Updation while Creating FTs

maneeSHA-256:
## Problem

`token_denom` — the counter `lockTokensForSplitOnce` reads to pick which
denominations to select — was left stale by two paths that move RBT out of Free,
so later selections asked for tokens that no longer were, failing with
`lockSelectedTokens: no tokens provided` and blaming whatever ran next.

## Fixes

- **FT mint denom** — `UpdateTokenInfo` incremented instead of decrementing, and
  keyed on a `TokenValue` never copied: a phantom `denom=0` row, real
  denomination missed.
- **`.dockerignore`** — `docker/data/` stopped matching after the harness moved.

## Tests

First `token_denom` invariant coverage: `ft_parts.py` (FTs minted from a parts-only wallet — the main FT suite mints
from `did_a`, whose whole tokens are selected first). Both auto-run under
`--run-all-tests`; `ft-parts` also gets a standalone CI job, because inside the
full matrix it skips itself once earlier phases have drained node A.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,7 +9,7 @@ mac/
 linux/
 windows/
 logs/
-docker/data/
+test/docker/data/
 
 # ---- Prebuilt binaries committed at repo root (rebuilt inside the image) ----
 /ipfs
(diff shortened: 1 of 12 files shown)


Allen-Cherian (approved):
Vaishnav CH tested and approved.

PR #741, opened 2026-09-03
Fix NFT/SC tokens left permanently Locked after a failed transaction on a subscribed NFT

### LINKED ISSUE #740: Child NFT creation stays broken after "Quorum is not setup" error
Attempting to create a child NFT without quorum configured returns the error Quorum is not setup. This is expected. However, after this failed attempt, child NFT creation stops working entirely — subsequent attempts fail as well.

Allen-Cherian:
## Problem

  A failed transaction on an NFT held by subscription left the NFT stuck in `Locked`.                                                                                                                     
  
  `BuildTransactionInfoFromRequest` locks NFT/SC tokens without an ownership filter and tags
  them with `lock_reference_id = reqID`. The failure-path release matched on `did = initiator`
  instead, so a subscribed NFT (whose `tokens.did` is the remote owner) was never released.

  Symptom: a child-mint that fails at pledge (e.g. `requestPledgeTokenHandler : Quorum is not
  setup`) leaves the parent `Locked`, and every later attempt fails with
  `not in executable status`, even after the quorum is fixed.

  ## Fix
  
  - Replace `ReleaseAllLockedNFTAndSCTokensForDID` with `ReleaseLockedNFTAndSCTokensByReference`,
    keyed on `lock_reference_id` (the same key the RBT and FT releases already use).
  - Restores `Deployed`/`Executed` by `latest_role` as before, clears the stale reference, and
    refuses an empty reference so it can never release unscoped.
  - No longer frees tokens locked by a concurrent request for the same DID.

Closes #740

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	version        string = "1.0.0"
+	version        string = "1.0.5SC5"
 	currentCommit  string = "unknown"
 	previousCommit string = "unknown"
 )
--- a/core/core.go
+++ b/core/core.go
@@ -269,6 +269,10 @@ func (c *Core) SetupCore() error {
 	if c.ipfsOps != nil {
 		c.w.SetIPFSOperations(NewWalletIPFSAdapter(c.ipfsOps))
 	}
+	// Release NFT/SC tokens left Locked by a crash or restart mid-transaction; the deferred cleanup in initiateTransaction never ran for them.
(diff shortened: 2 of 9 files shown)


arnabghose997 (dismissed):
Tested with the following scenario:

Restarted the quorum node and did not setup the Quorum. Proceeded to create a child NFT from a parent NFT, and the execution of parent NFT happened successfully.

gklps (approved):
LGTM

PR #743, opened 2026-09-10
fix(consensus): reject a DID acting as its own quorum

Allen-Cherian:


core/transaction.go
```diff
@@ -50,7 +50,36 @@ func (c *Core) initiateTransaction(reqID string, request *models.TransactionRequ
...
+	}
+	if len(quorumAddresses) == 0 {
+		c.log.Error("InitiateTransaction: No quorums available")
+		resp.Message = "InitiateTransaction: No quorums available for transaction"
+		return resp
+	}
+	c.log.Info("InitiateTransaction: Quorums found", "count", len(quorumAddresses), "primaryQuorum", quorumAddresses[0])
 
+	// A DID must not pledge for its own transaction: both roles would lock under one reference ID and the pledge handler would free the transfer tokens.
+	if _, quorumDID, _ := util.ParseAddress(quorumAddresses[0]); quorumDID == initiatorDID {
```
thewebchap, line 70: Should it be quorumAddresses[0] only or have to go through all?

core/transaction.go
```diff
@@ -50,7 +50,36 @@ func (c *Core) initiateTransaction(reqID string, request *models.TransactionRequ
...
+		c.log.Error("InitiateTransaction: initiator DID is configured as its own quorum", "did", initiatorDID)
+		resp.Message = "InitiateTransaction: initiator DID cannot act as its own quorum"
+		return resp
+	}
+
+	// The quorum must not be the receiver either; only ownership-moving requests name a receiver, so NFT execute is exempt.
+	movesOwnership := request.Tokens.RBT > 0 || len(request.Tokens.FT) > 0 || request.Tokens.TransferNFTOwnership
+	if _, quorumDID, _ := util.ParseAddress(quorumAddresses[0]); movesOwnership && nextOwnerDID != "" && nextOwnerDID != initiatorDID && quorumDID == nextOwnerDID {
```
thewebchap: why nextOwnerDID != initiatorDID is required?

core/consensus/checks.go
```diff
@@ -74,6 +74,105 @@ func ValidateTransactionInfoFields(txnInfo *models.TransactionInfo) error {
...
+
+	for _, quorum := range txnInfo.Quorums {
+		if quorum == nil {
+			continue
+		}
+		for _, t := range quorum.Tokens {
+			if t == nil || t.TokenID == "" {
+				continue
+			}
+			if role, clash := consumed[t.TokenID]; clash {
```
thewebchap, line 150: Will it be better to group by tokenvalue and do match check only within them? We can group by tokevalue looking at tokenID

thewebchap (changes requested):

core/transaction.go
Allen-Cherian: The idea was to not to allow self transfer. Do we need to remove this check ?

thewebchap: self transfer can happen . Also necessary in some situation like parts?

core/consensus/checks.go
gklps: I don't think grouping by value can change the result, and it would cost us NFT/SC coverage.

- Value is derived from the token ID itself (GetTokenValueFromTokenID), so two equal IDs always land in the same value bucket: grouping can't separate a collision or reveal a new one, it just adds a step.
- It would also cost us coverage: GetRbtIDElements rejects NFT/SC (Qm prefix) and FT IDs, so those tokens would have no bucket and we'd have to skip them (the exact bypass the check closes)

gklps: And the cost is already linear: one map from the transfer set, one hash lookup per pledge token, no pairwise comparison

core/transaction.go
Allen-Cherian: Understood. will remove that check.

core/transaction.go
Allen-Cherian: So right now all our cases are based on a single quorum perspective. Only the first quorum will be picked up in all cases.

thewebchap (approved):

PR #744, opened 2026-09-10
fix(consensus): return the stored signature when a quorum receives the same consensus request twice

### LINKED ISSUE #742: Consensus retry after a lost quorum reply forks the token chain
Summary

  When the owner's consensus request reaches the quorum but the reply is lost in transit, the PeerManager automatically re-sends the identical request. The quorum has already committed the block from the first attempt, so its chain-integrity check rejects the retry as a conflicting block. The owner then marks the transaction failed and never persists the block. From that point the owner and quorum hold different chain heads and every subsequent transaction on that token fails with a chain mismatch.

  Four earlier occurrences of this fork coincided with a panic in core/sync.go, which was assumed to be the cause. This occurrence had no panic, crash, or restart on either node. The panic is only one way to lose the reply. The underlying defect is that the consensus round is not idempotent on the quorum side.

  Environment

  - Testnet deployment, smart contract execution path (hasSC=true, transaction value 0.001)
  - Owner node and quorum node on separate VMs, connected through the libp2p tunnel on 127.0.0.1:22011
  - Same contract executed repeatedly by one initiator DID at roughly one execution per second

 Root cause

  1. The quorum commits the block and pledges before its HTTP reply reaches the owner. There is a window of tens of milliseconds where the quorum has state the owner does not.
  2. If the connection drops in that window, the owner sees EOF with no status and cannot tell whether the quorum committed.
  3. The PeerManager retries the same request. The quorum has no record that it already answered this transaction id, so it validates the block against its chain tip, sees that the tip is the same block, and rejects it.
  4. Nothing reconciles the two nodes afterwards. The owner's failed transaction is treated as never having happened.

Allen-Cherian:
## Problem
  
  When the initiator sends a consensus request, the quorum validates it, pledges,
  commits the block, and replies with its signature. If that reply is lost on the
  wire (we saw a libp2p stream reset returning EOF, but a sync panic or a restart
  has the same effect), the initiator's PeerManager retries the identical request.

  The quorum already has that block as its chain head, so TokenChainIntegrityCheck
  treats the retry as a conflicting block and rejects it with 400. The initiator
  then marks the transaction failed and never persists the block. From that point
  the initiator is one position behind the quorum, and every later transaction on
  that token is rejected with the same integrity error until an operator manually
  replays the missing block.

  This was seen on the YMCA reward contract on 9 Sep 2026 and in four earlier
  forks on the same deployment.

  ## Fix

  Before running full validation, initiateConsensusHandler looks up the incoming
  transaction id in the quorum's own transactions table. If a row exists carrying
  this quorum's signature, and that signature verifies against the incoming
  TransactionInfo, the handler returns the stored signature with status true and
  leaves the already-pledged tokens untouched. Any miss falls through to the
  existing validation path unchanged.

  The transaction id is a hash of the serialized TransactionInfo, so a retried
  request always carries the same id as the committed one, and a modified payload
  produces a different id and is validated normally.

  No initiator-side change is needed: the existing three-attempt retry in
  PeerManager now recovers the lost signature on its own.

  ## Verification

  On a local testnet node running as quorum:

  - Re-sending a consensus request the quorum had already committed returned 200
    with the identical stored signature. Token rows and tokenchain were unchanged
    and the abort-release guard did not fire.
  - The same request with a modified memo took the normal path and was rejected
    by full validation with 400.

  Fixes #742

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -269,6 +269,10 @@ func (c *Core) SetupCore() error {
 	if c.ipfsOps != nil {
 		c.w.SetIPFSOperations(NewWalletIPFSAdapter(c.ipfsOps))
 	}
+	// Release NFT/SC tokens left Locked by a crash or restart mid-transaction; the deferred cleanup in initiateTransaction never ran for them.
+	if _, err := c.w.ReleaseStaleNFTAndSCLocksOnStartup(c.Ctx); err != nil {
+		c.log.Error("Failed to release stale NFT/SC locks on startup", "err", err)
+	}
 	c.PingSetup()
 	c.CheckQuorumStatusSetup()
 	c.peerSetup()
--- a/core/quorum_initiator.go
+++ b/core/quorum_initiator.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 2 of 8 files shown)


gklps (approved):
LGTM

PR #751, opened 2026-09-21
test: remove legacy tests/ suite superseded by test/integration

Allen-Cherian:
Removes the root-level tests/ directory, the old Python suite that drove the node through CLI subcommands and flags (generatetestrbt, -grpcPort) that no longer exist in the binary; it has been untouched
  since the March Postgres migration and cannot run.
  The tests.yml workflow that invoked it was already removed on 2026-09-10, and the committed test/integration harness covers RBT, NFT, FT and smart-contract flows through the HTTP API, so nothing in the
  repo references the deleted tree.


gklps (approved):
LGTM

PR #752, opened 2026-09-22
Fix whole rbt minter check

maneeSHA-256:
## Summary

- Fixes part-token transfers failing `ValidateMinterAllowlist` after the first hop following a split. The receiver asked only the current sender for the whole-token genesis, and a later holder never has the burnt whole token's chain.
- Closes a gap where a peer could serve any genesis that names an allowlisted minter. The gate now checks that the genesis mints the requested whole token and is signed by the claimed minter.
- Cuts repeat network fetches by storing the burnt whole token's chain locally once it has been resolved.

## Changes

- `resolveSplitInitiator` reads the part token's own height-0 (split) transaction and asks that split's initiator for the genesis first. If that fails it falls back to the declared `genesisPeer`, the same peer is never asked twice, and the error lists every peer that failed.
- `genesisMintsToken`: the served genesis must mint `wholeID` with an empty `PreviousTransactionID`, otherwise it is rejected. This check always runs.
- `verifyFetchedGenesisSignature` / `Core.verifyGenesisSignature`: verifies the initiator signature against the minter's DID, using `InitialiseDID` to resolve it. A failure rejects the genesis only when `enforceGenesisSignature` is set.
- New `Core.SyncBurntTokenChainFromPeer`: saves the whole token's chain from the peer that served the genesis, but only if the chain's last entry is a burn. A failure here is only logged.
- `ValidateTransaction` / `ValidateMinterAllowlist` take two new callbacks (`syncBurntChain`, `verifyGenesisSig`), passed in from the quorum and fullnode.

## Testing

- New unit tests in `minter_allowlist_test.go` check peer order and fallback, no duplicate fetches, fullnode chain reads, when the chain is saved and that a save failure is non-fatal, rejection of genesis for another token or a part token, and signature verification and enforcement. `go test ./core/consensus/` passes.


core/consensus/minter_allowlist.go
```diff
@@ -33,9 +80,69 @@ func ValidateMinterAllowlist(
...
+	return validateMinterAllowlist(txnInfo, isFullnode, w, log, fetchGenesisTx, syncBurntChain, verifyGenesisSig, testnet, mainnet)
+}
+
+// enforceGenesisSignature gates whether a whole-token genesis whose signature
+// fails to verify is REJECTED (true) or merely logged (false).
+//
+// Staged rollout, deliberately off. Verification needs the claimed minter's DID
+// document, which Core.InitialiseDID may have to fetch over the network, so a
+// benign resolution failure would reject an otherwise-valid mainnet transfer.
+// Run log-only until telemetry shows no benign failures, then flip to true.
```
gklps: Where is it flipped to false? once its true when will it revert to false?

core/consensus/minter_allowlist.go
```diff
@@ -33,9 +80,69 @@ func ValidateMinterAllowlist(
...
+
+// verifyFetchedGenesisSignature checks that the genesis carries the claimed
+// minter's own signature over it.
+//
+// A whole-token mint sets Initiator == Owner == the minting DID and signs with
+// util.SignTransaction (Wallet.PersistGenesisTokenRecord), so the initiator
+// signature is exactly what util.VerifySignature validates. Passing this means
+// the attacker must hold an allowlisted minter's private key, not merely name
+// its DID.
+func verifyFetchedGenesisSignature(
```
gklps, line 130: Can we check the signature using original bytes and not just struct?

something like 

`
id := hex.EncodeToString(util.CalculateHash(genesisTx.Info, constants.HashAlgorithm_SHA3_256))
`
`
if id != genesisTx.ID {
    return fmt.Errorf("genesis bytes do not hash to its ID")
}
sigBytes, err := util.Base64ToBytes(sig.InitiatorSignature)
`
`
if err != nil {
    return err
}
ok, err := dc.SignVerify([]byte(genesisTx.ID), sigBytes)
`

core/consensus/minter_allowlist.go
maneeSHA-256: It was false and I changed it to true (line 99), missed updating the comments. Will update it.

core/consensus/minter_allowlist.go
maneeSHA-256: Ultimately, that it happening. Since there is already a function to verify the signature on transaction, I reused that function here. Original byte verification function, which is used here, is defined in util/transaction.go: 40

core/consensus/minter_allowlist.go
maneeSHA-256: updated

gklps (approved):
LGTM

